PR #1, opened 2017-06-15
Migration of sui-cz and sui mono 

davidbarna:
This is a migration from https://github.com/SUI-Components/commit-release-manager/
To review, as it's copied code, the best things is to either review commit (they're very small) by commit of compare last commit to 0cfc2deec125bd92c041db8200ab301ccb223fd3, which is the commit where I started modifying and adapting commit-release-manager copied code.

@naxhh , @carlosvillu, @miduga please review


package.json
```diff
@@ -7,7 +7,26 @@
     "monorepo",
     "sui"
   ],
-  "scripts": {},
+  "scripts": {
+    "commit": "sui-mono commit",
```
naxhh: We are using "co" everywhere. Maybe keep it as co? or standarize this everywhere :\

I personally do npm run co and if don't works I assume it does not have cz and use git commit.

packages/sui-cz/README.md
```diff
@@ -0,0 +1,27 @@
...
+npm install --save-dev validate-commit-msg
+npm install --save-dev @schibstedspain/sui-cz
+```
+
+in `package.json`
+
+```js
+{
+  "config": {
+    "validate-commit-msg": {
```
naxhh: Maybe we should add husky in the install + config ?

If not people will just copy this and think it should works.

package.json
```diff
@@ -7,7 +7,26 @@
...
+    "release": "sui-mono release",
+    "release:check": "sui-mono check",
+    "commitmsg": "validate-commit-msg"
```
naxhh: I understand husky runs this instead of having it on config?

👍

packages/sui-cz/README.md
```diff
@@ -0,0 +1,27 @@
+# sui-cz
+
+This is a commitizen adapter for the sui-mono tool and any other packages of sui family
+
+This is required to be in a different repo by cz, and the package does not use our `@schibsted` prefix because of how commitizen bootstrap resolves routes.
```
naxhh: This can be removed 😊

packages/sui-mono/README.md
```diff
@@ -0,0 +1,134 @@
+# Commmit Release Manager
```
naxhh: sui-mono and next line too.

package.json
davidbarna: Exactly. Husky adds "native" scripts to npm scripts correponding to git hooks
https://github.com/typicode/husky/blob/master/HOOKS.md

packages/sui-cz/README.md
davidbarna: Done.

Issue #13, opened 2017-06-21
[sui-mono] Make "build" script NOT mandatory


davidbarna:
### Package

sui-studio@1.1.0

### Description

Packages with no build command cannot be committed

### Steps to Reproduce

1. Create a new package without any script
2. Commit it with "sui-mono commit"
3. Publish with it with "sui-mono release"

**Expected behavior:** package is published

**Actual behavior:** package is not published and an error is thrown of missing "npm run build" command

**Reproduces how often:** always

### Additional Information

Executing build makes no sense. Developer should simply add a "prepublish" command that would be automatically executed by npm client.

Nevertheless, to be retrocompatible, we'll make `npm run build` to be executed **only if it exits**

PR #14, opened 2017-06-21
Bug #13

davidbarna:
@carlosvillu and @miduga please review


packages/sui-mono/package.json
```diff
@@ -1,15 +1,11 @@
 {
   "name": "@schibstedspain/sui-mono",
-  "version": "1.2.0",
+  "version": "1.5.0",
```
davidbarna: Upgrade of more than one minor for testing purpose. I wanted to try publishing with and without build script in "real"

PR #17, opened 2017-06-22
Migrate sui studio

davidbarna:
## Description
First step of migration.
check and relase commands were delegated to sui-mono

@carlosvillu and @naxhh please review


packages/sui-studio/bin/helpers/cli.js
```diff
@@ -0,0 +1,40 @@
...
+}
+
+/**
+ * Parses sui-mono config in package.json and returns corresponding CLI options
+ * @return {Array<String>} CLI options
+ */
+function getSuiMonoOptions () {
+  const path = require('path')
+  const pkg = require(path.join(__dirname, '..', '..', 'package.json'))
+  const conf = pkg && pkg.config && pkg.config['sui-mono']
```
naxhh: Doesn't make more sense to hardcode this values instead of taking configuration?

Since any `sui-studio` studio will use the same options?

packages/sui-studio/bin/sui-studio-init.js
```diff
@@ -0,0 +1,174 @@
...
+    "react": "15",
+    "react-dom": "15"
+  },
+  "dependencies": {},
+  "config": {
+    "ghooks": {
+      "pre-commit": "echo \\"Precommits...\\"",
+      "commit-msg": "validate-commit-msg"
+    },
+    "validate-commit-msg": {
```
naxhh: This should point to `sui-mono/types`

packages/sui-studio/bin/sui-studio-init.js
```diff
@@ -0,0 +1,174 @@
...
+        "fix",
+        "refactor",
+        "test",
+        "docs",
+        "release",
+        "chore",
+        "perf"
+      ]
+    },
+    "commitizen": {
```
naxhh: This and next section one are not required.

packages/sui-studio/bin/helpers/cli.js
davidbarna: I thought it was nice to make it taken from config... This way if you update sui-mono with new features, you only have to update the config naturally and not to remember the magic behaind sui-mono execution in sui-studio.

packages/sui-studio/bin/sui-studio-clean-modules.js
```diff
@@ -0,0 +1,21 @@
+/* eslint no-console:0 */
+
+const colors = require('colors')
+const {remove} = require('fs-extra')
+const BASE_DIR = process.cwd()
+const cwds = require('./walker').componentsFullPath(BASE_DIR)
```
davidbarna: Ok. Not scope of this PR.

packages/sui-studio/bin/sui-studio-init.js
```diff
@@ -0,0 +1,174 @@
...
+`${BASE_DIR}/${PROJECT_NAME}/package.json`,
+`{
+  "name": "${PROJECT_NAME}",
+  "version": "1.0.0",
+  "description": "",
+  "private": true,
+  "scripts": {
+    "phoenix": "rm -Rf node_modules && npm it",
+    "deploy": "suistudio build && surge public/ -d ${PROJECT_NAME}.surge.sh",
+    "co": "git-cz",
```
davidbarna: Not needed actually. Will be removed.

packages/sui-studio/bin/sui-studio-init.js
```diff
@@ -0,0 +1,174 @@
...
+  "devDependencies": {
+    "@schibstedspain/suistudio-fatigue-dev": "github:SUI-Components/suistudio-fatigue-dev"
+  },
+  "peerDependencies": {
+    "react": "15",
+    "react-dom": "15"
+  },
+  "dependencies": {},
+  "config": {
+    "ghooks": {
```
davidbarna: The init command will be moved to a new sui-studio-create.
sui-studio will be executed locally and sui-studio-create globally. I'll refactor in sui-studio-create directly.

packages/sui-studio/bin/helpers/cli.js
naxhh: I'm ok with that. but since everyone will have the same conf :P

packages/sui-studio/bin/sui-studio-init.js
naxhh: that sounds nice!

packages/sui-studio/bin/helpers/cli.js
davidbarna: Well, I've reverted these commits...
The options thing was getting more complicated when sui-cz was involved...
So, we'll do that sui-studio-create add "sui-mono" config in package.json. Otherwise, code would be messy.

Issue #10, opened 2017-06-21
[sui-studio-create] New package for studio creation


davidbarna:
### Package

sui-studio-create@1.0

### Description

Create a package to be executed globally, that does current behavior of `suistudio init <project_name>`

Issue #9, opened 2017-06-21
[sui-studio] Make it work locally


davidbarna:
### Package

sui-studio@latest

### Description

To be able to migrate common deps to sui-studio, we need to make it work locally.

**When** sui-studio is executed
**Then** it should check is there is a local and/or global instance of it
**When** there is a local instance of the studio, 
**Then** it should warn the user it's using local version and do it
**Otherwise** it should throw error and ask user to `npm install sui-studio --save-dev`

miduga:
Maybe we should separate the global `studio create` and the other studio commands. So the first is global, but the latter is only local.

PR #18, opened 2017-06-23
Consolidate new version of sui-studio to work locally

davidbarna:
## Description
* All commands were tested in local execution with bug fixing.
* sui-studio-create was created as a new package to execute globally
* sui-component-peer-dependencies was created as a package to include in all hosts of components. For instance, studios.
* "sui-studio run-all" now runs newly created  "sui-mono run"

**This is not final. sui-studio does not work locally yet. Plus, we have to merge changes made in original repo**

## Related Issue
#9 and #10 

@carlosvillu and/or @naxhh and/or @miduga have a look if you can.


packages/sui-mono/bin/sui-mono-run.js
```diff
@@ -0,0 +1,20 @@
+const path = require('path')
+const program = require('commander')
+const config = require('../src/config')
+const serialSpawn = require('../src/spawn').serialSpawn
+const PACKAGES_DIR = path.join(process.cwd(), config.getPackagesFolder())
+const cwds = config.getScopes().map(pkg => path.join(PACKAGES_DIR, pkg))
+
+program.parse(process.argv)
+
+const tasks = cwds.map(cwd => getTaskArray(cwd))
```
naxhh: If i'm not wrong tihs can be just `.map(getTaskArray)` but not really critical

packages/sui-mono/src/spawn.js
```diff
@@ -0,0 +1,55 @@
+/* eslint no-console:0 */
```
naxhh: I think I'm missing something, but what is the difference between this and `execute.js` ?

packages/sui-studio-create/src/index.js
```diff
@@ -0,0 +1,99 @@
...
+
+writeFile(
+`${BASE_DIR}/${PROJECT_NAME}/package.json`,
+`{
+  "name": "${PROJECT_NAME}",
+  "version": "1.0.0",
+  "description": "",
+  "private": true,
+  "scripts": {
+    "phoenix": "rm -Rf node_modules && npm i",
```
naxhh: Our current studio was created with this one:

```
"phoenix": "rm -Rf node_modules && npm i && suistudio clean-modules && suistudio run-all npm i",
```

Maybe this one makes more sense? since it will be applied to all components

PD: similar for linting commands

packages/sui-mono/src/spawn.js
davidbarna: I created it at first for sui-precommit and copied it there to have a single way to execute...
command execute have multiple ways all over the packages. Most of them trim colors and newlines of the output. But in this case execute does about the same thing.
Anyways serialSpawn was about to be moved to sui-helpers to execute commands from there and have a single way to to it.

packages/sui-mono/bin/sui-mono-run.js
davidbarna: Done

packages/sui-mono/src/spawn.js
naxhh: nice, don't care wat to use but, use just one 😄

davidbarna: YEs and this approach is more clear, documented and ready for executions of normal repos and mono repos.

PR #23, opened 2017-06-27
Make sui-studio@4 stable as local package

davidbarna:
## Description
sui-studio@4 seems to work locally.
Now suistudio-fatigue-deps is not need in anywhere. To create a studio and make it work, the only dep needed is at root level as you can see [here](https://github.com/SUI-Components/sui/blob/migrate-sui-bundler-from-sui-studio-webpack/packages/sui-studio-create/src/index.js#L81). The components don't need suistudio-fatigue-deps anymore to work as part of a studio.

As a PoC, the optimizely studio was migrated to sui-studio@4, with the following commits: https://github.com/SUI-Components/react-ab-testing/pull/1/commits . Please review this code too.

**Before making it official, I will create a branch at [mt--uilib-components](https://github.schibsted.io/scmspain/frontend-mt--uilib-components) and migrate it to studio-4, to make sure it's stable**

Repos left to migrate (secondary, I think):
* sui-decorators
* sui-i18n
* sui-react-domain-connector
* sui-ssr

## Related Issue
#9 

@miduga @naxhh @carlosvillu please review carefully


packages/sui-mono/bin/sui-mono-release.js
```diff
@@ -38,7 +38,7 @@ const releaseEachPkg = ({pkg, code} = {}) => {
       ['git', ['add', cwd]],
       ['sh', ['-c', `VERSION=$(node -p -e "require('./package.json').version") && git commit -m "release(${pkg}): v$VERSION"`]],
       ['npm', ['publish', `--access=${publishAccess}`]],
-      ['git', ['push', 'origin', 'master']]
+      ['git', ['push', 'origin', 'HEAD']]
     ]
     scripts['build'] && commands.unshift(['npm', ['run', 'build']])
```
naxhh: Now that I'm checking this.

and the migration of the optimizely studio, do you need at all the task `version: npm run build` in the component scripts?

davidbarna: You're right. Will be removed.

Issue #21, opened 2017-06-27
[sui-mono] CLI sui-mono link


davidbarna:
### Package
sui-mono@latest

### Description
Create new command that links packages between each other.
Use this code as inspiration: https://github.com/cyclejs/cyclejs/blob/master/.scripts/symlink-to-each-other.sh

PR #34, opened 2017-07-03
Issue/21 Command sui-mono link

davidbarna:
## Description
sui-mono link link all packages to each other (only if needed)

## Related Issue
#21 

Please review @miduga @carlosvillu @naxhh

### DIFF
--- a/packages/sui-helpers/packages.js
+++ b/packages/sui-helpers/packages.js
@@ -1,0 +1,116 @@
+const path = require('path')
+
+/**
+ * Get absolute paths of packages
+ * @param  {String} cwd
+ * @param  {Array} packages
+ * @return {Array}
+ */
+const getPackagesPaths = (cwd) => {
+  const getPath = path.join.bind(this, cwd)
+  return (packages) => packages.map((pkg) => getPath(pkg))
+}
+
+/**
+ * Get Package JSON info
+ * @param  {String} packagePath Absolute pathç
+ * @return {Object} {} in case of error
+ */
+const getPackageJson = (packagePath) => {
+  try {
+    return require(path.join(packagePath, 'package.json'))
+  } catch (e) {
+    return {}
+  }
+}
+
+/**
+ * Get list of dependencies of a path
+ * @param  {String} packagePath Absolute pathç
+ * @return {Array} Both prod and dev deps are returned
+ */
+const getPackageDependencies = (packagePath) => {
+  const pkg = getPackageJson(packagePath)
+  return [...Object.keys(pkg.dependencies || {}), ...Object.keys(pkg.devDependencies || {})]
+}
+
+/**
+ * Get npm names of packages from paths
+ * @param  {String} cwd
+ * @param  {Array} packages
+ * @return {Array}
+ */
+const getPackagesNames = (cwd) => {
+  const getPaths = getPackagesPaths(cwd)
+  return (packages) => {
+    return getPaths(packages)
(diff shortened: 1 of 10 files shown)


packages/sui-helpers/packages.js
```diff
@@ -0,0 +1,116 @@
+const path = require('path')
```
davidbarna: Let me know if you think that the code is crazy

packages/sui-helpers/packages.js
```diff
@@ -0,0 +1,116 @@
...
+ */
+const mapNameToPath = (cwds) => (packages) => (name) => cwds[packages.indexOf(name)]
+
+/**
+ * Get which dependencies are included in provided path
+ * @param  {Array<String>} dependencies NPM names of packages
+ * @param  {String} pkgPath Path to browse
+ * @return {Array<Array>} index[0] is a path, index[1] re found dependencies
+ */
+const mapDepedencies = (dependencies) => (pkgPath) => {
```
naxhh: More than map this should be filterDeps or getDependenciesBeingUsed or similar 😜

packages/sui-helpers/packages.js
```diff
@@ -0,0 +1,116 @@
...
+/**
+ * Get list of npm names of packages used locally by each other
+ * @param  {Array<String>} cwds Absolute paths of projects
+ * @param  {Array<String>} dependencies NPM names of packages
+ * @return {Array<String>} NPM names of packages
+ */
+const getUsedInternalDependencies = (cwds) => (dependencies) => {
+  return getInternalDependencyMap(cwds)(dependencies)
+    .reduce((result, [, dependencies]) => [...result, ...dependencies], []) // Reduce to array of dependencies
+    .filter((dep, idx, dependencies) => dependencies.indexOf(dep) === idx) // Remove duplicated
```
naxhh: maybe is worth moving this filter function out to `unique` or `filterUnique`

packages/sui-helpers/packages.js
```diff
@@ -0,0 +1,116 @@
...
+}
+
+/**
+ * Get path of a package from npm name
+ * @param  {Array<String>} cwds Absolute path of packages
+ * @param  {Array<String>} packages Names of packages
+ * @param  {String} name Package name
+ * @return {String}
+ */
+const mapNameToPath = (cwds) => (packages) => (name) => cwds[packages.indexOf(name)]
```
naxhh: ❤️

packages/sui-studio/bin/sui-studio-link.js
```diff
@@ -6,6 +6,8 @@ const fs = require('fs-extra')
 const program = require('commander')
 const BASE_DIR = process.cwd()
 
+console.log(colors.yellow('sui-studio link is deprecated. Use "sui-studio link-all" instead.'))
```
naxhh: Makes sense to deprecate link?

link-all will allow us to link all packages in the same studio, but it wont allow to link packages to other studios, while link does

packages/sui-helpers/packages.js
davidbarna: If I do this to the filter, then chaining would no be possible and this func is less readable.
If I abtract only the function, it would be somthing like `const filterUnique = (item, idx, items) => items.indexOf(item) === idx` and I don't like its params...

packages/sui-helpers/packages.js
davidbarna: Done.

Issue #39, opened 2017-07-04
[babel-preset-sui] Add legacy decorators to plugins


miduga:
### Package

[babel-preset-sui]@1

### Description

It's missing legacy decorators thus some packages are not working as expected.

### Steps to Reproduce

1. Try to compile with the preset a project with decorators.

**Expected behavior:** Compiles! :)

**Actual behavior:** It's not compiling! It throws an error! :(

PR #40, opened 2017-07-04
feat(babel-preset-sui): add legacy decorators plugin to make compatible some packages

miduga:
Add Legacy Decorators plugin.

ISSUES CLOSED: #39


packages/babel-preset-sui/package.json
```diff
@@ -12,6 +12,7 @@
     "babel-plugin-transform-async-generator-functions": "6.22.0",
     "babel-plugin-transform-async-to-generator": "6.22.0",
     "babel-plugin-transform-class-properties": "6.23.0",
+    "babel-plugin-transform-decorators-legacy": "1.3.4",
```
miduga: This should be after the `class-properties` one.

PR #44, opened 2017-07-10
fix(sui-bundler): use contenthash because chunkhash sometimes doesnt change

miduga:
Use contenthash instead chunkhash.


miduga:
:+1: dedo de Carlos

PR #49, opened 2017-07-14
Feature/mono repo mono package check release

naxhh:
## Description
Allow mono repo mono package to perform check and release commands

## Example

Since I've changed a bit how release works I also provide an example of multi-package

Keep in mind that for release the commands for testing where:

```
    commands = [
      ['echo', [`${RELEASE_CODES[code]}`]],
      ['echo', [`${cwd}`]],
      ['echo', [`${packageName}`]]
    ]
```

I've only performed a full release on a mono-package repo to confirm everything was working

### Mono repo multi package

**Check**

```
ignacio.tolstoy in ~/code/frontend-mt--uilib-components (master) $ ../sui-components/sui/packages/sui-mono/bin/sui-mono.js check
RELEASES TO DO

We checked all packages and recent commits, and discovered that
according to ComVer https://github.com/staltz/comver you should
release new versions for the following packages.

`feed/articles` needs a new MINOR version released because:
  . fix(feed/articles): always perform redux call

`navigation/breadcrumb` needs a new MINOR version released because:
  . feat(navigation/breadcrumb): pass link factory to the breadcrumb

`navigation/pagination` needs a new MINOR version released because:
  . feat(navigation/pagination): update prev/next links on page change
  . fix(navigation/pagination): make it render on server
```

**Release**

```
ignacio.tolstoy in ~/code/frontend-mt--uilib-components (master) $ ../sui-components/sui/packages/sui-mono/bin/sui-mono.js release

echo minor @feed/articles
minor

echo /Users/ignacio.tolstoy/code/frontend-mt--uilib-components/components/feed/articles @feed/articles
/Users/ignacio.tolstoy/code/frontend-mt--uilib-components/components/feed/articles

echo feed/articles @feed/articles
feed/articles

echo minor @navigation/breadcrumb
minor

echo /Users/ignacio.tolstoy/code/frontend-mt--uilib-components/components/navigation/breadcrumb @navigation/breadcrumb
/Users/ignacio.tolstoy/code/frontend-mt--uilib-components/components/navigation/breadcrumb

echo navigation/breadcrumb @navigation/breadcrumb
navigation/breadcrumb

echo minor @navigation/pagination
minor

echo /Users/ignacio.tolstoy/code/frontend-mt--uilib-components/components/navigation/pagination @navigation/pagination
/Users/ignacio.tolstoy/code/frontend-mt--uilib-components/components/navigation/pagination

echo navigation/pagination @navigation/pagination
navigation/pagination
```

### Mono package

**Check**

```
ignacio.tolstoy in ~/code/frontend-mt--lib-motor (master) $ ../sui-components/sui/packages/sui-mono/bin/sui-mono.js check
RELEASES TO DO

We checked all packages and recent commits, and discovered that
according to ComVer https://github.com/staltz/comver you should
release new versions for the following packages.

`@schibstedspain/motor` needs a new MAJOR version released because:
  . feat(fetcher): remove fake fetcher subscriptions
  . fix(editorial): fix server error string

```

**Release**

```
ignacio.tolstoy in ~/code/frontend-mt--lib-motor (master) $ ../sui-components/sui/packages/sui-mono/bin/sui-mono.js release

echo major @code/frontend-mt--lib-motor
major

echo /Users/ignacio.tolstoy/code/frontend-mt--lib-motor @code/frontend-mt--lib-motor
/Users/ignacio.tolstoy/code/frontend-mt--lib-motor

echo META @code/frontend-mt--lib-motor
META
```

Note: You may notice the `@code/frontend-mt-lib-motor`
That's because our command lib takes the "scope" from the PWD and in this case is the last 2 folders of the path. But we are publishing the package correctly as `@schibsted/...`


packages/sui-mono/bin/sui-mono-release.js
```diff
@@ -18,26 +17,34 @@ const RELEASE_CODES = {
   3: 'major'
 }
 
-const releasesByPackages = (pkg) => {
```
naxhh: I've changed how this function works because it made not much sense and would make the mono-package check a bit harder/duplicated.

Before we where requesting for each scope a check and retrieving only the part of the given scope
Now we just perform a check that will return all the scopes (only 1 in the case of monopackage) and work with that information.

It should make the release a bit faster but I don't think it will be noticed by the users

packages/sui-mono/src/config.js
```diff
@@ -35,6 +36,15 @@ module.exports = {
...
   getPublishAccess: function () {
     return publishAccess
+  },
+  getProjectName: function () {
+    return projectPackage.name
+  },
+  isMonoPackage: function () {
+    const folders = cwds(path.join(basePath, packagesFolder), deepLevel)
+    const scopePath = folders.pop()
+
```
carlosvillu: I don't have a better idea, but if it only uses the first folder to define when all the project is mono-package, it may be problematic.

But I don't have a better idea either. What do you think ?!

naxhh: we can check if several of them have a package.json is a multi-package.

But:
- Should we check all of them?
- If one does not have the package.json is a multi-package? 
- If non of them but one have it, is multi-package?

Maybe we should check all of them and all should have the package.json.

Since the folders will be a scope and if you try to release that scope it will fail?

carlosvillu: I would say that `If one does not have the package.json is a multi-package`

What do you think ?!

davidbarna: Maybe deeplevel should be 0 by default and we should check this . Then we would have to add deeplevel =1 to current mono repos... but not a big deal and the rest would make sense IMO

naxhh: @carlosvillu As we agreed I'll check all the scopes and if one does not have the package.json is not a multi package.

I just didn't have the time to do it

@davidbarna I'm not sure if I understood this.

If deeplevel = 0 we wouldn't have any scope. and we may still want scopes in a mono package repo.

We use 1 scope 1 package but in a mono package scopes are informative..

"I've made changes in the Url parsing module" kind of

davidbarna: Ok I get it. Nervermind.

naxhh:
Changes done.

- Fix how mono repo commits where flattened
- If one scope has no package.json is a monopackage repo

Issue #28, opened 2017-06-29
[sui-studio] Shall we run npm install after generating a component?


davidbarna:
Issue from previous repo
https://github.com/SUI-Components/SUIStudio/issues/19

julian-gernun:
+1

davidbarna:
I would put it as an option. Then you can create your "npm run generate" with that option if you want.

julian-gernun:
Can you think of any use case in which you wouldn't want to install after creating it?

PR #52, opened 2017-07-18
Feature/npm install after generating component

miduga:
Please review @carlosvillu @davidbarna 

## Description
Fixes https://github.com/SUI-Components/sui/issues/28
Also, it checks if the component directory exists in order to avoid generating a component that already exist.

### DIFF
--- a/packages/sui-studio/bin/sui-studio-generate.js
+++ b/packages/sui-studio/bin/sui-studio-generate.js
@@ -2,8 +2,10 @@
 
 const program = require('commander')
 const colors = require('colors')
+const fs = require('fs')
 const fse = require('fs-extra')
 const pascalCase = require('pascal-case')
+const spawn = require('child_process').spawn
 
 program
   .option('-R, --router', 'add routering for this component')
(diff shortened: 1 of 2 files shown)


packages/sui-studio/bin/sui-studio-generate.js
```diff
@@ -55,124 +57,140 @@ const COMPONENT_PLAYGROUND_FILE = `${DEMO_DIR}playground`
...
+      i18n: {t (s) { return s.split('').reverse().join('') }}
+    }
+  }
+  `
+  )
+])
+.then(() => {
+  console.log(colors.gray(`[${packageName}]: Installing the dependencies`))
+  const install = spawn('npm', ['install'], {cwd: COMPONENT_DIR})
+
```
davidbarna: FYI in sui-helpers there is a helper to to spawn that returns a promise and you won't have to handle output or errors.

miduga: :+1: Will rewrite then all the spawns on the studio with that. Thx!

davidbarna: Watch out ! The behavior may defer. spawn has shell:true by default.
I'd do it only for the command you are working on because it's a new one.

PR #55, opened 2017-07-20
fix(sui-studio-create): wrong package.json generation

davidbarna:
## Description
Fix traling comma un valid in package.json


packages/sui-studio-create/src/index.js
```diff
@@ -71,7 +71,7 @@ writeFile(
     "lint": "npm run lint:js && npm run lint:sass",
     "commitmsg": "validate-commit-msg",
     "precommit": "sui-precommit run",
-    "generate": "sui-studio generate -P re -S schibstedspain",
+    "generate": "sui-studio generate -P re -S schibstedspain"
```
naxhh: just noticed this, why sui-studio-creates uses the prefix re by default?

PR #57, opened 2017-07-21
release(sui-bundler): update dependencies and to webpack 3

miduga:
Please, review @carlosvillu and @davidbarna 

Update to version 3 of webpack, update dependencies, add ModuleConcatenationPlugin (but it's not working because the way we're using modules with babel, something that we have to check for sure).

Tested on Fotocasa build and is working as expected. :) We might need to test with sui-studio and coches bundle.

I'm publishing a major because while the API is the same it's a huge change so we offer a rollback.


miduga:
Updated to webpack 3.4

PR #62, opened 2017-07-25
[sui-bundler] happypack

carlosvillu:
[sui-bundler]

## Description
Add HappyPack to improve the build times


miduga:
Have you seen any improvement with that? :)

PR #65, opened 2017-08-01
fix(sui-studio): disabling on the fly loaded style

julian-gernun:
## Description
Disables the stylesheet created on Style component unmount

## Related Issue
All previous style versions where enabled having to reload the page in order to get rid of the removed css properties while developing a component

For example:

X component style file version 1:

```css
someSelector {
  margin: 5px;
  padding: 15px;
}
```

X component style file version 2:

```css
someSelector {
  padding: 20px;
}
```

As the styles are added on the fly via blobs, the browser was applying:

```css
someSelector {
  margin: 5px;
  padding: 20px;
} 
```

### DIFF
--- a/packages/sui-studio/src/components/style/index.js
+++ b/packages/sui-studio/src/components/style/index.js
@@ -19,8 +19,12 @@ export default class Style extends Component {
     this._linkElement = createLinkElement()
   }
 
+  componentWillUnmount () {
+    const lastIndex = document.styleSheets.length - 1
+    document.styleSheets[lastIndex].disabled = true
+  }
+
   render () {
-    // https://github.com/webpack-contrib/style-loader/blob/master/addStyles.js#L238
     const blob = new window.Blob([this.props.children], {type: 'text/css'})
     const oldSrc = this._linkElement.href
     this._linkElement.href = window.URL.createObjectURL(blob)


packages/sui-studio/src/components/style/index.js
```diff
@@ -19,8 +19,12 @@ export default class Style extends Component {
     this._linkElement = createLinkElement()
   }
 
+  componentWillUnmount () {
+    const lastIndex = document.styleSheets.length - 1
+    document.styleSheets[lastIndex].disabled = true
+  }
+
   render () {
-    // https://github.com/webpack-contrib/style-loader/blob/master/addStyles.js#L238
```
julian-gernun: 404

carlosvillu: The new one:
https://github.com/webpack-contrib/style-loader/blob/master/lib/addStyles.js

packages/sui-studio/src/components/style/index.js
```diff
@@ -19,8 +19,12 @@ export default class Style extends Component {
     this._linkElement = createLinkElement()
   }
 
+  componentWillUnmount () {
+    const lastIndex = document.styleSheets.length - 1
+    document.styleSheets[lastIndex].disabled = true
+  }
```
julian-gernun: Another approach would be to run through all styleSheets searching for the same href, but as far as I know nothing else injects stylesheet in between. Let me know if this could be dangerously.

carlosvillu: Maybe you can use an ID en the tag, to know always which one is the correct Style Tag. That solution can be a little fragile.

packages/sui-studio/src/components/style/index.js
carlosvillu: The new: https://github.com/webpack-contrib/style-loader/blob/master/lib/addStyles.js

packages/sui-studio/src/components/style/index.js
julian-gernun: Why? Do you think that href might change? As document.styleSheets is an array of CSSStyleSheet objects, which inherits from the StyleSheet object (see its [interface](https://drafts.csswg.org/cssom/#stylesheet)), we've a problem if the href might change!

carlosvillu: we are supposing that it is always the last, I right now it is. But that can change.

I mean, for me the PR it is OK. That will work for sure, the only thing is that with a ID, you will know always exactly with node is, without import it other element in the studio id add other styles tag too and changing the order.

Issue #67, opened 2017-08-02
[META] create a package.json inside each folder


carlosvillu:
### Package

[META]

### Description

In order to sui-mono detect this repo like a monorepo - multipackage, each folder inside `packages` must have a `package.json`

PR #69, opened 2017-08-02
Bug/meta add packagejson

carlosvillu:
## Description
Resolve issue #67 

@SUI-Components/developers Review please


naxhh:
if you don't set private:true they will be released on next release, is that expected?

Issue #60, opened 2017-07-24
[sui-domain-connector] Migrate ddd-react-redux to the module


carlosvillu:
### Package

[sui-domain-connector]

### Description

Migrate the package https://github.com/SUI-Components/ddd-react to the sui monorepo

PR #74, opened 2017-08-02
[sui-react-domain-connector] 🌈 Migrate package

carlosvillu:
ISSUES CLOSED: #60

## Description
Migrate legacy package


packages/sui-react-domain-connector/README.md
```diff
@@ -46,7 +35,7 @@ render(
 )
 ```
 
-Or passing store directly, if you need to use it outside the tree:
+O pasando directamente el store, si necesitas usarlo fuera del arbol de componentes de react:
```
miduga: Better english. :D

Issue #76, opened 2017-08-03
[sui-studio] API Tab error for some components


julian-gernun:
### Package

sui-studio

### Description
API Tab not working for some components

### Steps to Reproduce

1. Open http://mt-components.surge.sh/workbench/tracking/tealium/documentation/api
2. Open console
3. Click on "API" tab

**Expected behavior:**
It should render all props

**Actual behavior:**
<img width="669" alt="sui_studio" src="https://user-images.githubusercontent.com/17549662/28909131-62513cc8-7827-11e7-8aad-ca1b6d76f84b.png">

### Additional Information

Looks like it happens for every component that uses `withLocalService` from `@schibstedspain/ddd-react-redux`

PR #77, opened 2017-08-03
Search for component.js where there is one

carlosvillu:
ISSUES CLOSED: #76

## Description
When wrapped the component in a HoC we lost the API doc

## Related Issue
solved #76 

@SUI-Components/developers Review!

### DIFF
--- a/packages/sui-studio/src/components/documentation/try-require.js
+++ b/packages/sui-studio/src/components/documentation/try-require.js
@@ -4,18 +4,30 @@ const reqComponentsReadme =
   require.context(`bundle-loader?lazy!raw-loader!${__BASE_DIR__}/components`, true, /^\.\/\w+\/\w+\/README\.md?/)
 
 // https://webpack.github.io/docs/loaders.html#loader-order
-const reqComponentsSrc =
+const reqComponentsSrcIndex =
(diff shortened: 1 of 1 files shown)


packages/sui-studio/src/components/documentation/try-require.js
```diff
@@ -4,18 +4,30 @@ const reqComponentsReadme =
...
 const tryRequire = ({category, component}) => {
   const src = new Promise(resolve => {
     require.ensure([], () => {
       let bundler
-      try {
-        bundler = reqComponentsSrc(`./${category}/${component}/src/index.js`)
-      } catch (e) {
-        bundler = reqComponentsSrc(`./${category}/${component}/src/index.jsx`)
+
+      if (!bundler && reqComponentsSrcComponent.keys().includes(`./${category}/${component}/src/component.js`)) {
```
carlosvillu: I really need vacances. Any idea to improve this `ifs` ?!

miduga: I didn't try that, but something like that should work:

```
const regex = /\/src\/(component|index).jsx?$/g;
const bundler = reqComponentsSrcComponent.keys().find(key => key.match(regex))
return typeof bundler === 'undefined'
            ? console.error('No bundler found')
            : bundler(resolve)
```

carlosvillu: We have two context: reqComponentsSrcComponent and reqComponentsSrcIndex

PR #94, opened 2017-08-28
fix(sui-bundler): fix json files not loading in sass imports

davidbarna:
## Description
Webpack was loading sass-loader without its options (jsonImporter) for files outside of the context: sui-studio/src.
Created a plugin that ensure loader are instantiated with the same options in all

@naxhh @kikoruiz @miduga please review

### DIFF
--- a/packages/sui-bundler/plugins/loader-options.js
+++ b/packages/sui-bundler/plugins/loader-options.js
@@ -1,0 +1,30 @@
+/**
+ * Simple plugin that guarantees loaders to be loaded with the given options
+ * in any context.
+ * This fixes a bug of webpack that was loading files outside of context with
+ * the given loaders, but not their options.
+ *
+ * @param {Object} [options={}] key = loader name, value = loader options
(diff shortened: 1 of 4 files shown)


packages/sui-bundler/plugins/loader-options.js
```diff
@@ -0,0 +1,30 @@
...
+ * This fixes a bug of webpack that was loading files outside of context with
+ * the given loaders, but not their options.
+ *
+ * @param {Object} [options={}] key = loader name, value = loader options
+ */
+function LoaderUniversalOptionsPlugin (options = {}) {
+  this.loaderOptions = options
+}
+
+LoaderUniversalOptionsPlugin.prototype.apply = function (compiler, compilation) {
```
davidbarna: Don't know how to optimize it. Take into account the loader name is converted to loader path by webpack.
And the path is different denpending on context (sui-studio vs sui-bundler)

PR #98, opened 2017-08-30
fix(sui-mono): fix changelog to generate log in right context

davidbarna:
## Description
The command needs to be executed in the root of the repo, like release command.
Any command that uses isMonoPackage() func needs to be executed with right cwd as this functions depends on the cwd to determine if the current repo is either mono package or multipackage.

@miduga and @naxhh please review

### DIFF
--- a/packages/sui-mono/bin/sui-mono-release.js
+++ b/packages/sui-mono/bin/sui-mono-release.js
@@ -69,20 +69,26 @@ const releaseEachPkg = ({pkg, code} = {}) => {
     const pkgInfo = require(path.join(cwd, 'package.json'))
     const scripts = pkgInfo.scripts || {}
 
-    let commands = [
+    let releaseCommands = [
       ['npm', ['--no-git-tag-version', 'version', `${RELEASE_CODES[code]}`]],
       ['git', ['add', cwd]],
-      ['git', ['commit -m "release(' + packageScope + '): v$(node -p -e "require(\'./package.json\')".version)"']],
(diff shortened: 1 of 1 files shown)


packages/sui-mono/bin/sui-mono-release.js
```diff
@@ -69,20 +69,26 @@ const releaseEachPkg = ({pkg, code} = {}) => {
...
       ['git', ['tag -a ' + tagPrefix + '$(node -p -e "require(\'./package.json\')".version) -m \"v$(node -p -e "require(\'./package.json\')".version)\"']], // eslint-disable-line no-useless-escape
+      scripts['build'] && ['npm', ['run', 'build']],
       !pkgInfo.private && ['npm', ['publish', `--access=${publishAccess}`]],
       ['git', ['push', '--tags', 'origin', 'HEAD']]
     ].filter(Boolean)
-    scripts['build'] && commands.unshift(['npm', ['run', 'build']])
 
-    serialSpawn(commands, {cwd})
+    serialSpawn(releaseCommands, {cwd})
+      .then(() => serialSpawn(docCommands))
```
naxhh: Maybe I don't understand the problem.

This is because we want to generate the CHANGELOG for all component on the root package?

davidbarna: We want to generate the log for one package, but with the monorepo as cwd.
This way, when isMonoPackage is executed, it would return false, not true.

PR #102, opened 2017-08-31
Issue/29 Fix for travis deploys

davidbarna:
## Description
Fix for linux systems (travis-ci)
As now-cli copies url to clipboard, I was get it from there to create the alias.
But now-cli does not copy the deployment url on linux
Travis build was breaking on deployment. It took so long to find an alternative solution until I found out there was the simplest: https://github.com/zeit/now-client
I don't deserve my salary today...

@miduga @naxhh please review


packages/sui-studio/bin/sui-studio-deploy.js
```diff
@@ -31,26 +26,27 @@ if (typeof program.name !== 'string') {
...
+const now = new NowClient(process.env.NOW_TOKEN)
 
 getSpawnPromise('now', ['rm', deployName, '--yes', '-t $NOW_TOKEN'])
   .catch(() => {}) // To bypass now rm error on the first deploy
   .then(() => getSpawnPromise(
     'now', ['./public', '--name=' + deployName, '--static', '-t $NOW_TOKEN']
   ))
-  // Obtain deployment url copied to clipboard by now
-  .then(res => clipboardy.read())
+  .then(() => now.getDeployments())
```
miduga: `.then(now.getDeployments)`

naxhh:
Since you are using now client, doesn't make sense to move all the other now commands to the client? or is just not possible?

davidbarna:
@naxhh I thought about it and I didn't for two reasons:
1. Working as a CLI, you see the commands to reproduce manually if needed, which was really usefull until now.
2. The client has a different API then the CLI. For instance, getDeployments does not allow to send a name of deployments, deleteDeployment expects an id, not a name, and createDeployment is way more complicated as using the client than the "now deploy" command.
(3. Third non acceptable reason: not spend more time on this)

Issue #92, opened 2017-08-23
Prepare deployments for now.sh instead of surge


davidbarna:
Prepare deployments for now.sh instead of surge

PR #103, opened 2017-09-05
fix(sui-studio): fix: SPA links don't work in deployments

davidbarna:
## Description
SPA was browseable in the frontend but inner links could not be shared to open a component directly

## Related Issue
#92

@miduga and @carlosvillu please review

### DIFF
--- a/packages/sui-studio/bin/sui-studio-deploy.js
+++ b/packages/sui-studio/bin/sui-studio-deploy.js
@@ -4,6 +4,18 @@
 const program = require('commander')
 const NowClient = require('now-client')
 const { getSpawnPromise } = require('@schibstedspain/sui-helpers/cli')
+const {writeFile, removeFile} = require('./helpers/file')
+const BUILD_FOLDER = './public'
+const PKG_FILE = BUILD_FOLDER + '/package.json'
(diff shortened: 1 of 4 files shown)


packages/sui-studio/bin/sui-studio-deploy.js
```diff
@@ -36,9 +48,11 @@ const now = new NowClient(process.env.NOW_TOKEN)
 
 getSpawnPromise('now', ['rm', deployName, '--yes', '-t $NOW_TOKEN'])
   .catch(() => {}) // To bypass now rm error on the first deploy
+  .then(() => writePackageJson(deployName)) // Add package.json for SPA server
   .then(() => getSpawnPromise(
-    'now', ['./public', '--name=' + deployName, '--static', '-t $NOW_TOKEN']
+    'now', [BUILD_FOLDER, '--name=' + deployName, '--npm', '-t $NOW_TOKEN']
```
carlosvillu: That work with private repos ?!
I think that is better deploy and static not relative `npm` pages

davidbarna: It has to be for SPA as mentionnned by https://zeit.co/blog/now-static
Private package will be implemented later.

PR #107, opened 2017-09-06
sui-deploy package

davidbarna:
## Description
Move "sui-studio deploy" command to a new "sui-deploy" package, so the same logic can be used to deploy other SPAs

* [x] Move sui-studio helps to sui-helpers to be used by sui-deploy
* [x] Create sui-deploy for SPAs
* [x] Refactor sui-studio to use sui-deploy (preserve API)

## Reviews
@miduga @carlosvillu


packages/sui-deploy/README.md
```diff
@@ -0,0 +1,69 @@
...
+
+It provides:
+* Common way to deploy our projects
+* SPA deployment in now.sh
+
+
+## Installation
+
+```sh
+$ npm install @s-ui/sui-deploy --save-dev
```
carlosvillu: @s-ui ?!?!?!

packages/sui-deploy/bin/sui-deploy-spa.js
```diff
@@ -0,0 +1,64 @@
...
+  program.outputHelp()
+}
+
+if (!process.env.NOW_TOKEN) {
+  console.log('ERR: NOW_TOKEN env variable is missing')
+  process.exit(1)
+}
+
+const [,, deployName, folder = BUILD_FOLDER] = process.argv
+const now = new NowClient(process.env.NOW_TOKEN)
```
carlosvillu: Are There a way to avoid this spawn, or the NowClient ?!

Now, you are mixing the now binary install in your system and the now cliente library. I thing that could be great, is have now install in your system is not a requirement to use this package.

What do you think ?!

packages/sui-deploy/README.md
davidbarna: We decided it a while ago !

packages/sui-deploy/bin/sui-deploy-spa.js
davidbarna: We can't: https://github.com/SUI-Components/sui/pull/102#issuecomment-326507914

Issue #122, opened 2017-10-03
[babel-preset-sui] env preset


davidbarna:
### Package

[babel-preset-sui]@latest

### Description

Use env preset instead of plugins
Modifie browsers config according to SUI requirements

carlosvillu:
To use Env, we need a browser support matrix defined isnt ?!

davidbarna:
We'll apply fotocasa matrix. I'll ask them.
Basically latest chrome and firefox, IE11. In case of doubt we'll be restrictive. It can't be more restrictive than what we already have.

PR #128, opened 2017-10-04
feat(babel-preset-sui): replace es2015 preset by env

davidbarna:
ISSUES CLOSED: #122

No improvement of final build... Even trying to build only for Chrome it doesn't seem to be worth having separated bundles by env.

### DIFF
--- a/packages/babel-preset-sui/lib/index.js
+++ b/packages/babel-preset-sui/lib/index.js
@@ -3,14 +3,25 @@ const cleanList = require('./clean-list')
 
 module.exports = {
   'presets': cleanList([
-    require('babel-preset-es2015'),
+    ["env", {
+      "debug": false,
+      "targets": {
+        "node": "6.0.0",
+        "browsers": [
+          "> 1%",
+          "last 4 versions",
+          "Firefox ESR",
+          "Safari >= 6",
+          "iOS >= 7",
+          "not ie < 11"
+        ]
+      }
+    }],
     isInstalled(['preact', 'react'], 'babel-preset-react'),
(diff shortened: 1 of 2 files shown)


packages/babel-preset-sui/lib/index.js
```diff
@@ -3,14 +3,25 @@ const cleanList = require('./clean-list')
...
+      }
+    }],
     isInstalled(['preact', 'react'], 'babel-preset-react'),
     isInstalled('flow-bin', 'babel-preset-flow')
   ]),
   'plugins': [
-    require('babel-plugin-syntax-trailing-function-commas'),
     require('babel-plugin-transform-async-generator-functions'),
-    require('babel-plugin-transform-async-to-generator'),
     require('babel-plugin-transform-decorators-legacy').default,
```
davidbarna: We still have a lot of plugins because babel-preset-env@1 is es2015 based.
babel-preset-env@2 is not released yet.

packages/babel-preset-sui/lib/index.js
```diff
@@ -3,14 +3,25 @@ const cleanList = require('./clean-list')
 
 module.exports = {
   'presets': cleanList([
-    require('babel-preset-es2015'),
+    ["env", {
+      "debug": false,
+      "targets": {
+        "node": "6.0.0",
+        "browsers": [
+          "> 1%",
```
carlosvillu: That is the FC browser support matrix, isnt ?!

davidbarna: it is. I only added the node field. The rest is the same.

PR #130, opened 2017-10-06
feat(sui-mono): commit-all command

davidbarna:
## Description
commit-all command

### DIFF
--- a/packages/sui-mono/bin/sui-mono-commit-all.js
+++ b/packages/sui-mono/bin/sui-mono-commit-all.js
@@ -1,0 +1,75 @@
+#!/usr/bin/env node
+/* eslint no-console:0 */
+const program = require('commander')
+const path = require('path')
+const { exec } = require('child_process')
+const config = require('../src/config')
+const { getSpawnPromise, showError } = require('@s-ui/helpers/cli')
+
+program
+.usage('-m "My commit message"')
+.option('-m, --message "<message>"', 'Commit message')
+.option('-t, --type "<commit-type>"', 'Commit type')
+  .on('--help', () => {
+    console.log('  Description:')
+    console.log('')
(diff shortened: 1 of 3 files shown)


packages/sui-mono/bin/sui-mono-commit-all.js
```diff
@@ -0,0 +1,75 @@
...
+const packagesDir = path.join(process.cwd(), config.getPackagesFolder())
+
+/**
+ * Checks if given path has changes
+ * @param  {String}  path Folder to check
+ * @return {Promise<Boolean>}
+ */
+const hasChangedFiles = (path) => {
+  return new Promise((resolve, reject) => {
+    exec(`git add . && git status ${path}`, {cwd: path}, (err, output) => {
```
julian-gernun: I think you should at least advice that this command is going to commit everything

davidbarna: it's a specific command ti commit all... I don't get your point.
Plus, it's a local thing, a "git reset" is easy. There is no push, publish or anything with side effect.
What I can do, is document it more to explain more clearly what it does. But I really don't think it's necessary.

packages/sui-mono/bin/sui-mono-commit-all.js
```diff
@@ -0,0 +1,75 @@
...
+program
+.usage('-m "My commit message"')
+.option('-m, --message "<message>"', 'Commit message')
+.option('-t, --type "<commit-type>"', 'Commit type')
+  .on('--help', () => {
+    console.log('  Description:')
+    console.log('')
+    console.log('    Do same commit for all packages')
+    console.log('')
+    console.log('    Usefull for cross-pacakge refactors.')
```
julian-gernun: typo package

PR #131, opened 2017-10-06
Maintenance/Improvements

davidbarna:
## Description
* [x] feat(sui-cz): filter scopes without staged files
        Now the folder of commit is automatically selected, no need to use arrows.
* [x] fix(sui-studio-create): remove wrong tabs in package.json
* [x] feat(sui-studio-create): update with new deploy method
   Using sui-deploy istead of directly surge CLI
* [x] fix(sui-studio): generate deps without "latest"

### DIFF
--- a/packages/sui-cz/index.js
+++ b/packages/sui-cz/index.js
@@ -1,12 +1,28 @@
 /* eslint-disable no-console */
 
 const fs = require('fs')
+const path = require('path')
+const { exec } = require('child_process')
 const editor = require('editor')
 const temp = require('temp').track()
 const buildCommit = require('./buildCommit')
 const config = require('./config')
 
+const packagesDir = path.join(process.cwd(), config.getPackagesFolder())
 const scopes = config.getScopes().sort().map(name => ({name}))
+
+/**
+ * Check if modified files are present
+ * @param  {[type]}  path Folder to check
+ * @return {Promise<Boolean>}
+ */
+const hasChangedFiles = (path) => {
+  return new Promise((resolve, reject) => {
+    exec(`git status ${path}`, {cwd: path}, (err, output) => {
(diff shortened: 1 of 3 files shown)


packages/sui-cz/index.js
```diff
@@ -45,11 +61,17 @@ module.exports = {
             return [{name: 'META'}]
           }
 
-          if (typesWithOtherScopes.indexOf(answers.type) > -1) {
-            return scopes.concat(otherScopes)
-          }
-
-          return scopes
+          return Promise.all(scopes.map(pkg => (
```
davidbarna: We can also use the trick of the staged files to make sui-lint to execute only on changed files.
It would be WAY faster. What do you think?

packages/sui-studio-create/src/index.js
```diff
@@ -74,43 +74,44 @@ Promise.all([
...
+  "scripts": {
+    "phoenix": "rm -Rf node_modules && npm i && sui-studio clean-modules && sui-studio run-all npm i",
+    "deploy": "sui-deploy ${PROJECT_NAME} --now",
+    "co": "sui-studio commit",
+    "lint:js": "sui-lint js",
+    "lint:sass": "sui-lint sass",
+    "lint": "npm run lint:js && npm run lint:sass",
+    "commitmsg": "validate-commit-msg",
+    "precommit": "sui-precommit run",
+    "generate": "sui-studio generate -P re -S schibstedspain"
```
julian-gernun: Check this please.

packages/sui-cz/index.js
julian-gernun: I guess they can't be lint related integration issues, so 👍

davidbarna: Solved ! Thanks!

PR #149, opened 2017-11-09
feat(sui-component-peer-dependencies): bump to React 16

carlosvillu:
Bump React version to 16


miduga:
Mmmm I don't feel like react-helmet should be a peer dependency.

miduga:
LGTM! 🔝

PR #150, opened 2017-11-14
Include sui-js package

JoanClaret:
## Description
This package will contain a set of js utilities.
The first one will be a browser detection (device, browser, OS...).

## Usage
```sh
$ npm install @s-ui/js --save-dev
```

```js
import browser from '@s-ui/js/lib/browser'
const ua = browser(userAgent)
domain.config('isMobile', ua.mobile) // bool
```


miduga:
You don’t need babel-core nor babel-loader devDependencies.

JoanClaret:
i wasn't sure @miduga . well seen!

packages/sui-js/package.json
```diff
@@ -0,0 +1,27 @@
...
+  "scripts": {
+    "lib": "rm -Rf ./lib && mkdir -p ./lib && babel --presets sui ./src --out-dir ./lib",
+    "prepublish": "npm run lib"
+  },
+  "keywords": [],
+  "author": "Joan Claret <dpam23@gmail.com>",
+  "dependencies": {
+    "bowser": "1.8.1"
+  },
+  "license": "ISC",
```
davidbarna: MIT License I think...

packages/sui-js/README.md
```diff
@@ -0,0 +1,17 @@
...
+$ npm install @s-ui/js --save-dev
+```
+
+## browser
+A browser detector (device, browser, OS...).
+Imports [https://github.com/lancedikson/bowser](https://github.com/lancedikson/bowser).
+
+
+```js
+import browser from '@s-ui/js/lib/browser'
```
davidbarna: Does it make sense to name it browser if its a bowser thing ?
I mean if we add somday axios I assume we'll name it axios, not fetcher, isn't it ?

davidbarna:
Are we creating this repo because of the bowser need ???

I don't see the point right now...

before
```js
import bowser form 'bowser'
bowser._detect(navigator.userAgent)
domain.config('isMobile', bowser.mobile)
```
after
```js
import browser from '@s-ui/js/lib/browser'
const ua = browser(navigator.userAgent)
domain.config('isMobile', ua.mobile)
```

Plus, implicitly we are pretending that we can have independent instances of browser service, but bowser is used as singleton...

packages/sui-js/README.md
JoanClaret: from my point of view, today we are using bowser but tomorrrow we can use any other package. since this is browser related, i would keep using this name.

JoanClaret:
@davidbarna we are creating this package in order to use a cross teams solution to manage browser properties. like so, we provide an api to detect browser features. This is only a wrapper so we can change the `bowser` dependency in the future if we find another one with better performance or less weight. maybe it wasn't clear on the summary, sorry for that!

packages/sui-js/package.json
JoanClaret: You're right. thank you!

packages/sui-js/README.md
```diff
@@ -0,0 +1,17 @@
+# sui-js
+> Set of useful js utilities
+
+```sh
+$ npm install @s-ui/js --save-dev
```
kikoruiz: This is a PROD dependency, not DEV. So it should be just `--save`.

packages/sui-js/README.md
miduga: It makes sense as it's decoupled of the library we're using. We might use tomorrow another library or even create our own behind this name. That would be the idea of this package, being a black box that, with exports an API to get the browser info, but the developer doesn't need to know what's happening under the hood.

If someday we use axios, we will do a ' @s-ui/js/lib/fetcher'.

packages/sui-js/README.md
miduga: `$ npm install @s-ui/js --save` -> it should be for production.

JoanClaret: right. Thanks!

JoanClaret: right. Thanks!

packages/sui-js/README.md
davidbarna: I agree with the principle. But usually when you do that you create a façade with only the methods you wanna expose, not full access to a thirdparty API. This way if you change the library, you know exactly what you need to maintain. 
That's why I was asking. Either you expose expressively a thirdparty lib, either you limit the coupling to what you expose.

JoanClaret: Makes sense. Let me refactor some code and we'll take a look again.

JoanClaret: Done. Take a look @davidbarna

packages/sui-js/src/browser/index.js
```diff
@@ -0,0 +1,11 @@
+import bowser from 'bowser'
+
+const browser = (userAgent) => {
```
carlosvillu: I think that is API is too strict and hard to involved.
I would prefer put this behind an Object.
In this first version:
```js
export default {
  stats: au
}
```
And use it in this way:
`import {stats} from '@s-ui/js/browser'`
And later be apply put more helper with touch the public API

```js
export default {
  stats: au,
  cookies: [What_ever_cookie_parser_go_here]
}
```
And use it:
`import {stats, cookies} from '@s-ui/js/browser'`

JoanClaret: Done @carlosvillu . Check it out!

PR #154, opened 2017-11-17
feat(sui-js): adding js-cookie top sui-js

JoanClaret:
## Description
Parse, get and set cookies. Returns an object `cookie` with `parse`, `get` and `set` methods.

**Note:** `set` method does not work on server side.

```js
import cookie from '@s-ui/js/lib/cookie'

// Parse
const {parse} = cookie
domain.config('cookie', parse(cookies))

// Get
const {get: getCookie} = cookie
const smartBannerCookie = getCookie('smartbanner')

// Set
const {set: setCookie} = cookie
const setSmartBannerCookie = setCookie('smartbanner', 1)
```

### DIFF
--- a/packages/sui-js/README.md
+++ b/packages/sui-js/README.md
@@ -14,3 +14,25 @@ const {isMobile, osName} = stats(userAgent)
 domain.config('isMobile', isMobile) // bool
 domain.config('osName', osName) // string
 ```
+
+
+## cookie
+Parse, get and set cookies. Returns an object `cookie` with `parse`, `get` and `set` methods.
+
+**Note:** `set` method does not work on server side.
+
+```js
+import cookie from '@s-ui/js/lib/cookie'
+
+// Parse
+const {parse} = cookie
+domain.config('cookie', parse(cookies))
+
+// Get
+const {get: getCookie} = cookie
+const smartBannerCookie = getCookie('smartbanner')
+
+// Set
+const {set: setCookie} = cookie
+const setSmartBannerCookie = setCookie('smartbanner', 1)
+```
--- a/packages/sui-js/package.json
+++ b/packages/sui-js/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@s-ui/js",
-  "version": "1.0.0",
+  "version": "1.2.0",
   "description": "Set of JS utilities",
   "scripts": {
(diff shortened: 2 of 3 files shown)


packages/sui-js/src/cookie/index.js
```diff
@@ -0,0 +1,14 @@
+import {parse as parseCookie} from 'cookie'
+import jsCookie from 'js-cookie'
```
carlosvillu: I know that i had this conversation. Need we have this two libraries ?!

JoanClaret: Not sure about that. i think this two libraries has different features. i guess we can look for another one which includes all features we need.

JoanClaret: i could not find a repo with all features we need. i guess that's because parse is a different concept than get and set (even if they are closely related).
let's work with that!

packages/sui-js/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@s-ui/js",
-  "version": "1.1.0",
+  "version": "1.0.0",
```
kikoruiz: Is this a mistake?

JoanClaret: yes, it is! thank you!

packages/sui-js/README.md
```diff
@@ -14,3 +14,25 @@ const {isMobile, osName} = stats(userAgent)
...
+Parse, get and set cookies. Returns an object `cookie` with `parse`, `get` and `set` methods.
+
+**Note:** `set` method does not work on server side.
+
+```js
+import cookie from '@s-ui/js/lib/cookie'
+
+// Parse
+const {parse} = cookie
+domain.config('cookie', parse(cookies))
```
davidbarna: I guess what you need is to provide all cookies to the domain as an object.
Could you do that with js-cookie package using `Cookies.get()` ?

Or do you really need to actually parse the cookie hash somewhere ?

JoanClaret: looks like `Cookies.get()` could work fine, thanks for point it out!
let me make sure it works properly in server side. if so, we'll be able to get rid of `cookie` package and only use `jsCookie`. i'll keep you posted.

JoanClaret: @davidbarna unfortunatelly, `jsCookie` does not work in server side. so we need to load these two libraries.

PR #155, opened 2017-11-17
feat(sui-studio): exported data accesible from playground

julian-gernun:
With this changes a studio package can export whatever it wants and it will be defined in the playground file

### DIFF
--- a/packages/sui-studio/src/components/demo/index.js
+++ b/packages/sui-studio/src/components/demo/index.js
@@ -39,17 +39,24 @@ const cleanDisplayName = displayName => {
   return name !== undefined ? name : fallback
 }
 const pipe = (...funcs) => arg => funcs.reduce((value, func) => func(value), arg)
+const removeDefaultContext = exports =>
+  Object.keys(exports)
+  .filter(key => key !== DEFAULT_CONTEXT)
+  .reduce((acc, key) => {
+    acc[key] = exports[key]
+    return acc
+  }, {})
 
 export default class Demo extends Component {
   static bootstrapWith (demo, {category, component, style, themes}) {
-    tryRequire({category, component}).then(([Base, playground, ctxt, routes, events, pkg]) => {
+    tryRequire({category, component}).then(([exports, playground, ctxt, routes, events, pkg]) => {
       if (isFunction(ctxt)) {
         return ctxt().then(context => {
-          demo.setState({playground, Base, ctxt: context, routes, style, themes, events, pkg})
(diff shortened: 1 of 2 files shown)


packages/sui-studio/src/components/demo/index.js
```diff
@@ -39,17 +39,24 @@ const cleanDisplayName = displayName => {
   return name !== undefined ? name : fallback
 }
 const pipe = (...funcs) => arg => funcs.reduce((value, func) => func(value), arg)
+const removeDefaultContext = exports =>
+  Object.keys(exports)
+  .filter(key => key !== DEFAULT_CONTEXT)
+  .reduce((acc, key) => {
+    acc[key] = exports[key]
+    return acc
+  }, {})
```
carlosvillu: you could use a ES7 way:
```
const remove = exports => {
  const {DEFAULT_CONTEXT, ...rest} = exports
  return rest
}
```

packages/sui-studio/src/components/demo/index.js
```diff
@@ -150,7 +159,7 @@ export default class Demo extends Component {
 
         <Preview
           code={playground}
-          scope={{React, [`${cleanDisplayName(Enhance.displayName)}`]: Enhance, domain}}
+          scope={{React, [`${cleanDisplayName(Enhance.displayName)}`]: Enhance, domain, ...nonDefaultExports}}
```
carlosvillu: Are you try what happen when `nonDefaultExports ` is undefined ?!

packages/sui-studio/src/components/demo/try-require.js
```diff
@@ -43,6 +43,7 @@ const tryRequire = ({category, component}) => {
         const bundler = reqComponentsPlayGround(`./${category}/${component}/playground`)
         bundler(playground => resolve(playground))
       } catch (e) {
+        const Component = exports.default
         return resolve(`return (<${Component.displayName || Component.name} />)`)
```
julian-gernun: this isn't working and it wasn't working as exports/Component is undefined at this point. Is it ok if I change it to a console.error instead? Looks like this is executed if the playground can't be loaded

packages/sui-studio/src/components/demo/index.js
julian-gernun: this does not work

packages/sui-studio/src/components/demo/index.js
julian-gernun: yes, it still works

PR #156, opened 2017-11-22
feat(sui-js): add string utilities

kikoruiz:
## Description
Add a bunch of string utilities (`sui-js`). For now: `removeAccents` and `hasAccents` methods. Use npm `remove-accents` third-party package.

## Example
```js
import { removeAccents, hasAccents } from '@s-ui/js/lib/string'

console.log(removeAccents('París')) // "Paris"
console.log(hasAccents('Árbol')) // true
```

Please review @SUI-Components/developers !

### DIFF
--- a/packages/sui-js/src/string/index.js
+++ b/packages/sui-js/src/string/index.js
@@ -1,0 +1,4 @@
+import accents from 'remove-accents'
+
+export const hasAccents = accents.has
+export const removeAccents = accents.remove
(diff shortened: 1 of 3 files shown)


packages/sui-js/src/string/index.js
```diff
@@ -0,0 +1,4 @@
+import accents from 'remove-accents'
```
carlosvillu: Avoid import the library completaly. Try to require only the two functions

PR #158, opened 2017-11-28
feat(sui-domain): create sui-domain package

miduga:
A package to help you creating a domain in Schibsted Spain.


packages/sui-domain/scripts/postpublish.js
```diff
@@ -0,0 +1,17 @@
...
+
+const handleError = err => { if (err) throw err }
+
+fs.readdir(CWD, (err, files) => {
+  handleError(err)
+
+  files.forEach(file => {
+    const ext = path.extname(file)
+    EXTENSIONS_TO_REMOVE.includes(ext) &&
+      fs.unlink(path.join(CWD, file), handleError)
```
carlosvillu: I dont feel confortable removing file and copy files using CWD. I mean, we are totally sure that we dont remove users´s things ?!

Only to avoid put an extra folder in the require path

packages/sui-domain/src/fetcher/factory.js
```diff
@@ -0,0 +1,6 @@
+import AxiosFetcher from './AxiosFetcher'
```
davidbarna: Something to think about working with factories.
Having a single factory for all strategies forces the import of all used libraries.
For instance, imagine what will happen if we add the superAgentFetcher method, the code would be:

```js
import AxiosFetcher from './AxiosFetcher' // (imports axios package to bundle)
import SuperAgentFetcher from './SuperAgentFetcher' //(imports superagent package to bundle)

export default class FetcherFactory {

  static axiosFetcher = ({config}) =>
    new AxiosFetcher({config})

  static superAgentFetcher = ({config}) =>
    new SuperAgentFetcher({config})
}
```

Is it scalable ? Now that we are redefining part of the architecture, to me it makes sense to define separated factories.

Does it make sense ?

packages/sui-domain/src/index.js
```diff
@@ -0,0 +1,24 @@
...
+    this._config = config
+    this._useCases = useCases
+  }
+
+  get (key) {
+    if (key === 'config') {
+      return this._config
+    }
+
+    return this._useCases[key] || {
```
davidbarna: Why are we informing of the absence of the use case only on "execute" call?
It's very weird to me to return an actual object and not be able to find out if the use case is present on the "domin.get()" function call...
That's a homemade pattern and I don't get what is it for.

packages/sui-domain/src/fetcher/Fetcher.js
```diff
@@ -0,0 +1,36 @@
+/**
+ * @interface
+ * */
+export default class Fetcher {
```
davidbarna: As we use the "required" helper, it could be a good idea to do it with interfaces as well.

```js
const getMethods = obj => Object.getOwnPropertyNames(obj)
  .filter(key => typeof obj[key] == 'function')

const getClassName = obj => obj.name || `${obj.constructor.name}.prototype`

const compareMethods = compared => base => getMethods(base)
    .forEach((key) => {
      if(!compared[key] || base[key] === compared[key]) {
        throw new Error(`${getClassName(base)}.${key}() must be implemented.`)
      }
    })
   
const impl = instance => Class => {
   compareMethods(instance)(Class.prototype)
   compareMethods(instance.__proto__.constructor)(Class)  
}

export default impl
```

```js

import impl from './implements'
class Fetcher {
  constructor() {
     impl(this)(Fetcher)
  }
}
```

packages/sui-domain/scripts/postpublish.js
davidbarna: Same here.

miduga:
Sorry guys @carlosvillu @davidbarna. This PR was meant to request feedback and not to merge instantly. I don't know why the labels weren't attached. :D I will read your comments!

packages/sui-domain/src/index.js
miduga: I understand that this would keep your code working safely but breaking in the `get` won't as you're trying to chain the `get` method. I copied it as we're doing in the actual domains and might be @carlosvillu has a better explanation.

packages/sui-domain/src/index.js
```diff
@@ -0,0 +1,24 @@
+
```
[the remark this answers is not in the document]
miduga: 👍  will do.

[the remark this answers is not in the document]
miduga: I thought to not export for now the `required` method but if we find it useful we could recheck that.

packages/sui-domain/scripts/postpublish.js
miduga: We have three options:

1. Just export everything as an object in order to do:
`import { Entity, ValueObject } from '@s-ui/domain'`

2. Fix scripts and use `__dirname` instead `cwd`, as it points to the file location and it's safer.

3. Just keep the creepy `lib` stuff. The reason why I don't support this is not only aesthetic is as a concepto. If you has to explain a developer to import just the ValueObject, you has to explain why they have to put the `lib` in the require path everytime. Also, it feels like it's an internal stuff of the library while it's not, it's the main core of the package. I would like to avoid this pattern as much as possible but if there's not another good enough solution I will assume it.

packages/sui-domain/src/Fetcher.js
```diff
@@ -0,0 +1 @@
+export * as FetcherFactory from './fetcher/factory'
```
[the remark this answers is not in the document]
miduga: We might have to do, yes.

packages/sui-domain/src/fetcher/Fetcher.js
miduga: I would like to use Flow for this instead of having to put in every constructor we use. Even if we do not do this kind of check in production it feels we're going to far with that.

packages/sui-domain/src/fetcher/factory.js
miduga: I understand the Fetcher should be always a single implementation of a HttpFetcher. I agree that the naming couldn't be correct. I don't like to export this with a coupled library name as I don't want to let people to use whatever library they feel like (Axios, Superagent, fetch, petch, $.ajax,...). We should converge in the fetcher with the same library and then push forward that one.

packages/sui-domain/scripts/postpublish.js
davidbarna: I think that the priority is a clean public API. I really like the '@s-ui/domain/Entity'-like imports.
What about simply making more secure the prepublish and postpublish?
On postpublish we could do a safe delete: instead or traversing files in root dir you could read the files in lib and only delete in root what exist in lib and have the same content.
This way you get rid of @carlosvillu and have your flat package XD

packages/sui-domain/src/fetcher/factory.js
davidbarna: As I was saying it's more about defining the way to do a factory for domain more than for this package.
But if we define a way we should apply it here.

packages/sui-domain/scripts/postpublish.js
miduga: hahahahaha Sounds good. :D

packages/sui-domain/src/fetcher/factory.js
miduga: TODO: Rename `static axiosFetcher` for `static httpFetcher`.

packages/sui-domain/scripts/postpublish.js
miduga: We will use: `import { Entity, ValueObject } from '@s-ui/domain'

packages/sui-domain/src/index.js
miduga: Chequear si hacer el get del key esta es `undefined`.

PR #159, opened 2017-11-28
[sui-test] Create package

carlosvillu:
The idea behind this package is uniformate the way in the rest of teams run their tests.

- [x]  Provide a CLI
- [x] Replace all tests relative dev deps from the projects only for this one
  - [ ] WiP in motor with [this PR](https://github.schibsted.io/scmspain/frontend-mt--lib-motor/pull/240)


miduga:
It looks amazing! Let's transform this to a test-runner for any project and make sure it works and it would be ready to go! 👍

packages/sui-test-domain/src/describers/index.js
```diff
@@ -0,0 +1,3 @@
+import {describeOnServer, describeOnClient} from './serverDescribers'
+
+export {describeOnServer, describeOnClient}
```
naxhh: Remember to change this to auto detect node instead of rely on the package.json config

carlesnunez:
Maybe would be useful to have a README.md with info about usage of the tool

packages/sui-test/bin/karma/config.js
```diff
@@ -0,0 +1,65 @@
+var TARGET = process.env.npm_lifecycle_event
```
miduga: Why using `var` instead `const`?

packages/sui-test/bin/sui-test-browser.js
```diff
@@ -0,0 +1,40 @@
...
+const program = require('commander')
+const chalk = require('chalk')
+const runner = require('./karma')
+const { cleanStack } = require('./karma/util')
+
+program
+  .option('-W, --watch', 'Run in watch mode')
+  .option('-C, --ci', 'Run a Firefox headless for CI testing')
+  .on('--help', () => {
+    console.log('  Description:')
```
miduga: Instead bunch of console.log, maybe is a good option to use a template string here.

packages/sui-test/package.json
```diff
@@ -0,0 +1,33 @@
...
+  "version": "1.0.0-beta.10",
+  "description": "",
+  "bin": {
+    "sui-test": "./bin/sui-test.js"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@s-ui/helpers": "1.5.0",
```
miduga: We might want to use the major here instead the fixed version?

packages/sui-test/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "babel-core": "6.26.0",
+    "babel-polyfill": "6.26.0",
+    "babel-preset-sui": "1.5.0",
```
miduga: Same. In this case I think is particularly important to be consistent with the code we're testing.

packages/sui-test/package.json
carlosvillu: 👍

packages/sui-test/package.json
carlosvillu: 👍

packages/sui-test/bin/karma/config.js
carlosvillu: This is a legacy copy/pasted code. :)

PR #171, opened 2017-12-19
feat(sui-widget-embedder): added remoteCdn option on build command to…

carlesnunez:
# Remote CDN option
## Description
This pr adds the feature to implement the remoteCdn command that allows us build our sw.js pointing to the CDN URL that we desire.

## Example
```
sui-widget-embeder build -R http://mycdnurl.com
```


carlesnunez:
Done!

PR #172, opened 2017-12-22
feat(sui-react-domain-connector): add static method getInitialProps w…

frmedina:
…hen defined in component.

This PR is created to include the static method getInitialProps when is defined in the base component.

## Description
When trying to use the sui-react-domain-connector in the integration of New Construction Detail page in fotocasa web server, we found a problem that didn't let us to get anything retrieved from the static method getInitialProps (as it is not set in the mapUIServiceToProps) in our component's render function.
After some debugging, we realized that the method was not recovered from the base class, so we have included it in all files.

### DIFF
--- a/packages/sui-react-domain-connector/src/mapConfigToProps.js
+++ b/packages/sui-react-domain-connector/src/mapConfigToProps.js
@@ -4,6 +4,7 @@ import PropTypes from 'prop-types'
 const mapConfigToProps = (...configs) => Target => {
   class DDDConfigInjector extends Component {
     static displayName = `mapConfigToProps(${Target.displayName})`
+    static getInitialProps = Target.getInitialProps
     static contextTypes = {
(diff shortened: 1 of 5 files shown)


miduga:
Not really sure about if it's a good idea to couple our domain-connector to a specific solution for retrieving the initialProps. I mean, you could put the `getInitialProps` to the component resulted of the domain connector. That should work without touching the domain connector.

```
const NewConstructionDetailWithServices = withLocalService('whatever_use_case')(NewConstructionDetail)

NewConstructionDetailWithServices.getInitialProps = () => {}

export default NewConstructionDetailWithServices
```

frmedina:
@miduga I have just tried the workaround that you mention above, and it does work without changing the domain connector classes.
So, I will keep your suggested solution and close this PR.
Thanks for the advise!

PR #176, opened 2017-12-27
Feature/add md5 to sui js

miduga:
- [x] Add md5 as createHash to sui-js in order to be reusable across our projects.
- [x] Use smaller md5 implementation.
- [x] Remove md5 helper from sui-decorators and use new sui-js hash library.


packages/sui-js/src/hash/index.js
```diff
@@ -0,0 +1,5 @@
...
+const md5=function(c){var e,g,f,a,h=[];c=unescape(encodeURI(c));for(var b=c.length,k=[e=1732584193,g=-271733879,~e,~g],d=0;d<=b;)h[d>>2]|=(c.charCodeAt(d)||128)<<8*(d++%4);h[c=16*(b+8>>6)+14]=8*b;for(d=0;d<c;d+=16){b=k;for(a=0;64>a;)b=[f=b[3],(e=b[1]|0)+((f=b[0]+[e&(g=b[2])|~e&f,f&e|~f&g,e^g^f,g^(e|~f)][b=a>>4]+(m[a]+(h[[a,5*a+1,3*a+5,7*a][b]%16+d]|0)))<<(b=[7,12,17,22,5,9,14,20,4,11,16,23,6,10,15,21][4*b+a++%4])|f>>>32-b),e,g];for(a=4;a;)k[--a]=k[a]+b[a]}for(c="";32>a;)c+=(k[a>>3]>>4*(1^a++&7)&15).toString(16);return c} // eslint-disable-line
+
+export const createHash = md5
```
carlesnunez: Why don't just export createHash variable instead of declare it as md5 and equal to createHash for the export?

miduga: Well, because I've copied the md5 method from the repository of the first line. We might want to use a different algorithm in the future or even have two different `createHash` methods (createHashSecure` using another algorithm more secure but slower.) but I still would like to right now the one we're using. :D

kikoruiz:
@miduga Should this util behave like this one https://github.schibsted.io/scmspain/frontend-mt--web-app/blob/master/server/ssr/index.js#L86? If so, we can use this better ;)

miduga:
It should be doing exactly the same but with another API.

packages/sui-js/src/hash/index.js
nazarenomoreno: te amo midu

PR #182, opened 2018-01-12
feat(sui-component-peer-dependencies): update sui peer dependencies to latest React and React Router

miduga:
Updating sui-component-peer-dependencies to latest React and React Router.

Specially important to be able to use new <Fragment> API https://reactjs.org/docs/react-api.html#reactfragment.

React Router 3 is needed as the 2 version is not longer compatible with React 16.


miduga:
Dupe https://github.com/SUI-Components/sui/pull/149

PR #183, opened 2018-01-15
[sui-studio] Fullscreen mode

carlosvillu:
fullscreen

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "co": "sui-mono commit",
     "release": "sui-mono release",
     "release:check": "sui-mono check",
-    "lint": "sui-lint js && sui-lint sass",
+    "lint": "sui-lint js --fix && sui-lint sass",
     "wiki:pull": "git subtree pull --prefix=.wiki https://github.com/SUI-Components/sui.wiki.git master --squash",
     "wiki:push": "git subtree push --prefix=.wiki https://github.com/SUI-Components/sui.wiki.git master",
     "commitmsg": "validate-commit-msg",
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "co": "sui-mono commit",
     "release": "sui-mono release",
     "release:check": "sui-mono check",
-    "lint": "sui-lint js --fix && sui-lint sass",
+    "lint": "sui-lint js && sui-lint sass",
     "wiki:pull": "git subtree pull --prefix=.wiki https://github.com/SUI-Components/sui.wiki.git master --squash",
     "wiki:push": "git subtree push --prefix=.wiki https://github.com/SUI-Components/sui.wiki.git master",
     "commitmsg": "validate-commit-msg",
--- a/packages/sui-studio/src/components/demo/index.js
+++ b/packages/sui-studio/src/components/demo/index.js
@@ -1,7 +1,6 @@
 /* eslint react/no-multi-comp:0, no-console:0 */
 
 import PropTypes from 'prop-types'
-
 import React, { Component } from 'react'
 
 import {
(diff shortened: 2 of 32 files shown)


package.json
```diff
@@ -12,7 +12,7 @@
     "co": "sui-mono commit",
     "release": "sui-mono release",
     "release:check": "sui-mono check",
-    "lint": "sui-lint js && sui-lint sass",
```
davidbarna: If you do this, the sui-precommit will make changes without adding it to stage...
So the commit will be made with incorrect code.

packages/sui-studio/src/components/demo/index.js
```diff
@@ -139,34 +174,73 @@ export default class Demo extends Component {
...
 
   handleCode = () => {
-    this.setState({isCodeOpen: !this.state.isCodeOpen})
+    this.setState({ isCodeOpen: !this.state.isCodeOpen })
+  }
+
+  handleFullScreen = () => {
+    this.setState({ isFullScreen: !this.state.isFullScreen }, () => {
+      const { isFullScreen } = this.state
+      const aside = document.querySelector('.sui-Studio-sidebar')
```
davidbarna: Prefer getElementsByClassname over querySelector when selectors are classnames for performance concerns.

packages/sui-studio/src/components/demo/index.js
```diff
@@ -139,34 +174,73 @@ export default class Demo extends Component {
...
+      const toolBar = document.querySelector('.sui-StudioWorkbench-navigation')
+      const secundaryToolBar = document.querySelector(
+        '.sui-StudioNavBar-secondary'
+      )
+
+      if (isFullScreen) {
+        aside.classList.add('hidden')
```
davidbarna: Always better to add a single class once that will have several instructions.
For instance add ".sui-Studio--fullscreen" to the main element and hide the subelements by css.

package.json
carlosvillu: Yep!  I will revert it. And look for a better solución.

packages/sui-studio/src/components/demo/index.js
carlosvillu: Oki

packages/sui-studio/src/components/demo/index.js
carlosvillu: That brake the componetization way of think. Isn’t ?!  Because I display a component depending of my context in the page.

kikoruiz: You won't if you make component selectors depending on the global "classname", so that in your CSS will have something like this:
```css
.sui-Studio--fullscreen {
  & .aside-classname { display: none; }
}
```
And the same for the other components. Isn't it @davidbarna?

davidbarna: @carlosvillu to me what breaks the "componetization way of think" its to access remote components and alter their DOM model adding classnames.
I'm just saying to do it by css like @kikoruiz illustrated. It's less intrusive.

My main concern is performance. AFAIK, adding classes several times provokes several renders of the browser. It may be optimised as it's in the same call stack. But I it was I would still applied the mention solution. It'd still be less code and a much clearer approach to me.

carlesnunez:
Why are we getting from DOM the elements to set classNames on our sidebar? We could pass the method handleFullScreen from the topView and set 'isFullScreen' as a root state property so... pass it through property to our sidebar component and set its class in react-way.

davidbarna:
@carlesnunez that's a good point !
Make the demo to notify with an onFullScreen callback and make the parent (layout) handle the classes.
👍

carlosvillu:
David or Miguel can you make the changes and close the PR  please. I don’t know when I will came back to work. And is important close this PR.

davidbarna:
@carlesnunez what you mention cannot be done.
@kikoruiz and @miduga please review the ls commit with my changes.

PR #184, opened 2018-01-16
[sui-js] Created string functions fromSnakeToCamelCase and fromCamelToSnakeCase

carlesnunez:
## Description
Those functions have the purpose of transform our strings from this:
````
items_list_of_user
````
To this:
````
itemsListOfUser
````
And vice versa

### CODE REVIEW

File: packages/sui-js/src/string/index.js [hunk 1]
```diff
@@ -1,4 +1,2 @@
-import accents from 'remove-accents'
```
JoanClaret: should we remove it from dependencies?

davidbarna: Don't think so. It's used on line 2  :-)
```
export { has as hasAccents, remove as removeAccents } from 'remove-accents'
```

carlesnunez: Is exactly as @davidbarna said. We still using that dependency

Issue #163, opened 2017-12-03
[sui-domain] Avoid anemic models in the domain


carlosvillu:
```
class Father {
  constructor (params) {
    this._params = params
    Object.keys(params).forEach(key => (this[`_${key}`] = params[key]))
    console.log('constructor', params)
  }
  
  toJSON() {
    console.log('toJSON', this._params, Object.keys(this))
    return Object.keys(this).reduce((json, privateKey) => {
      const key = privateKey.replace('_', '')
      if (!this._params[key]) { return json }
      
      if (typeof this._params[key] === 'ValueObject') {
        return {...json, ...this._params[key].toJSON()}
      }
      
      json[key] = this._params[key]
      return json
    }, {})
  }
}

class Child extends Father {
  red = 1
  blue () {}
}

console.log(
	new Child({a:1, b:2}).toJSON()
)
```

davidbarna:
How? What's your proposal ?

PR #187, opened 2018-01-17
feat(sui-domain): created anemicModel to avoid unuseful code on our e…

carlesnunez:
## Description
This PR implements the AnemicModel on our sui-domain. Using this we will inherit extending from our entities and value object the properties with the convention 'this._propertyName' and the method .toJSON that transforms our AnemicModel into JSON valid object.

The **.toJSON** method will:
    * 1 - replace key _myPropertyKey by myPropertyKey
    * 2 - Check if the property was existing on the base properties constructor object or was added later. If second it will not included on te toJSON
    * 3 - Check if the property is AnemicObject and call its toJSON method too to get their JSON aspect.
    * 4 - Check if is an array, if it is will check if is an AnemicObject and do the same as step 3
    * 5 - If any of the latest steps were successful it will return the property without the dash.

#163

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,6 +13,12 @@
     "release": "sui-mono release",
     "release:check": "sui-mono check",
     "lint": "sui-lint js && sui-lint sass",
+    "test": "NODE_ENV=test npm run test:client && NODE_ENV=test npm run test:server",
+    "test:ci": "NODE_ENV=test ./node_modules/.bin/karma start --single-run --browsers Firefox",
+    "test:client": "NODE_ENV=test karma start",
+    "test:client:watch": "npm run test:client -- --no-single-run",
+    "test:server": "NODE_ENV=test mocha ./packages/*/test --require babel-core/register --recursive",
+    "test:server:watch": "npm run test:server -- --watch",
     "wiki:pull": "git subtree pull --prefix=.wiki https://github.com/SUI-Components/sui.wiki.git master --squash",
     "wiki:push": "git subtree push --prefix=.wiki https://github.com/SUI-Components/sui.wiki.git master",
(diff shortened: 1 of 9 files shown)


carlesnunez:
If I understood you this is not going to happen ever for the check that I do on the point 3:
https://github.com/SUI-Components/sui/pull/187/files#diff-7b683a16b375ade8a82d8e6f20f7d2faR20

carlesnunez:
Okay! I'll do that!

packages/sui-decorators/test/browser/cacheSpec.js
```diff
@@ -11,7 +11,7 @@ describe('Cache', () => {
   })
 
   describe('should be able use others method in this instance', () => {
-    it('return the response of other instance´s method', () => {
+    xit('return the response of other instance´s method', () => {
```
julian-gernun: What will happen to all these test you are skipping?

carlesnunez:
I wasn't able to fix it due to a unknowledge of these package. I've talked with @miduga with that and we will fix ASAP.

package.json
```diff
@@ -22,8 +28,25 @@
...
-    "validate-commit-msg": "2.12.2"
+    "karma": "2.0.0",
+    "karma-browserify": "5.1.2",
+    "karma-chrome-launcher": "2.2.0",
+    "karma-clear-screen-reporter": "1.0.0",
+    "karma-coverage": "1.1.1",
+    "karma-firefox-launcher": "1.0.1",
+    "karma-mocha": "1.3.0",
+    "karma-spec-reporter": "0.0.31",
+    "mocha": "^4.0.1",
```
julian-gernun: you could remove the ^

PR #193, opened 2018-01-26
Fix/fixed sui decorator test

carlesnunez:
## Description
The next PR fix partially the pending test with 'xit' on sui-decorator. 

What was happening with those tests was that on the new karma conf we've taking to run also the 'server' folder as karma valid tests. As we need to run those test on the server side the test were failing.

On the other hand we have the reverse case of the browse tests that were ran with mocha and should be ran by karma.

The strange thing here is that avoiding my errors on the implementation of the test runner in fact we have at least 2 server test that were not working even if they're ran by the test runner configured on its own package.

I've fixed one that were expecting an error object and we were passing a string with the error type.

## Pending to check

The other one is this one:
`NodeTracker must track to the server pass 20 seconds from the last track`

I'm not able to track what is happening here as my knowledge in this field is bad. Someone should help on that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -15,9 +15,9 @@
     "lint": "sui-lint js && sui-lint sass",
     "test": "NODE_ENV=test npm run test:client && NODE_ENV=test npm run test:server",
     "test:ci": "NODE_ENV=test ./node_modules/.bin/karma start --single-run --browsers Firefox",
-    "test:client": "NODE_ENV=test karma start",
+    "test:client": "NODE_ENV=test karma start --single-run",
     "test:client:watch": "npm run test:client -- --no-single-run",
-    "test:server": "NODE_ENV=test mocha ./packages/*/test --require babel-core/register --recursive",
+    "test:server": "NODE_ENV=test mocha './packages/*/test/!(browser)/*.js' --require babel-core/register --recursive",
     "test:server:watch": "npm run test:server -- --watch",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -15,9 +15,9 @@
     "lint": "sui-lint js && sui-lint sass",
     "test": "NODE_ENV=test npm run test:client && NODE_ENV=test npm run test:server",
```
julian-gernun: I'm not sure about this approach, what about every package manage its tests by itself? what about `sui-mono run npm test` in order to run all tests?

davidbarna: Agree with sui-mono test 100%. You should create the task if carles cannot do it.
About executing separately I do not. The logic is already lint by repo and executing are supposed to be isolated so executing all at once has better performance, and no drawback I think.
You could still execute sui-mono test in a separate test.

carlesnunez: IMHO is a fact that we avoid that some test were broken due to what you are proposing. I mean, if we need to handle X parallel processes and 1 karma.conf each time that we create a new packages... I'm not sure if it is the best approach. What are the reasons that make you say that @julian-gernun?

On the other hand... what are we going to do with the test that is failing? @miduga @davidbarna @julian-gernun

davidbarna: @carlesnunez it's DX I guess. To be able to work on one single package isolately

PR #203, opened 2018-02-12
fix(sui-helpers): fixed bug related to colors prototype inheritance b…

carlesnunez:
## Description
This PR fixes two strange behaviors related with the usage of sui-helpers/cli.

1. The no existence of the colors prototype properties on String javascript class due that the string that we are passing to the showError function once instantiated doesn't have all the prototype modifications needed to have the 'red' property. This was throwing the next error with an undefined print on the end that is not on the screenshot: 
captura de pantalla 2018-02-12 a las 19 36 33

This undefined print is the problem itself. Why its undefined? Because wen console.log evals the msg.red the 'red' property is undefined, this is caused because when msg was 'instantiated' String prototype wasn't modified yet by colors library.

With the fix this is what we will see:
captura de pantalla 2018-02-12 a las 19 33 10

2. The no co-relation between our commander instance and the error origin commander. We've been instantiating a new commander that doesn't match with the commander that throws the error. So, when the commander execute the outputHelp function it shows a default message:
captura de pantalla 2018-02-12 a las 19 25 42

If we pass the commander instance to the function and execute the outputHelp method we will see the helper information related with who executed the commander:
captura de pantalla 2018-02-12 a las 19 32 33

### DIFF
--- a/packages/sui-helpers/cli.js
+++ b/packages/sui-helpers/cli.js
@@ -1,9 +1,9 @@
 /* eslint no-console:0 */
-require('colors')
 const processSpawn = require('child_process').spawn
-const program = require('commander')
 const CODE_OK = 0
 const log = console.log
+const colors = require('colors')
+
 
 /**
  * Spawn several commands in children processes, in series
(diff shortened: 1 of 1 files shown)


carlesnunez:
@davidbarna I'm not really sure if it was a bug or a bad implementation from the widget-embedder side as I assume that we are using the show errors on other sui packages and we've never seen any issue.

davidbarna:
Very weird because all strings in js have access to the prototype AFAIK:

```js
String.prototype.myVar = 'hello'
"my string".myVar

// OUTPUT : 'hello'
```

carlesnunez:
I thought the same because prototype is shared. The other approach is that the two files have different 'worlds' of the js interpreter and different String implementations itself... 👎

carlesnunez:
In any case if we check their repo there are a couple of issues talking about the problems that their have using colors in the 'prototype' way. 😞 

https://github.com/Marak/colors.js/issues?utf8=%E2%9C%93&q=is%3Aissue+prototype+

PR #212, opened 2018-02-22
feat(sui-cz): made the main refactor and added the new commit saver

carlesnunez:
## Description
1. Made a refactor of the main index.js in order to increase de maintainability of the tool.
2. Implemented the 'commit saver' which is a sort of code that avoids us to lose our previous commit changes in case that the previous commit was not lint compliant or we've got an unexpected error(not added files, for example)
## Related Issue
N/A
## Example
sui-cz


packages/sui-cz/PrompterManager.js
```diff
@@ -0,0 +1,188 @@
...
+    })
+  }
+
+  /**
+   * Check if modified files are present
+   * @param  {[type]}  path Folder to check
+   * @return {Promise<Boolean>}
+   */
+  static checkIfHasChangedFiles (path) {
+    return new Promise((resolve, reject) => {
```
naxhh: I think we have a utils somewhere to perform exec that already converts to promises for you

packages/sui-cz/PrompterManager.js
```diff
@@ -0,0 +1,188 @@
...
+  static doCommit (commitString) {
+    ErrorCommitSaver.initErrorListener(commitString)
+    this.commit(commitString)
+    ErrorCommitSaver.discardOldCommit()
+  }
+
+  /**
+   * The get commmit steps method will return an array of objects that corresponds to the different steps of our commit promptet.
+   * [Step1, ...2, ...3, ...4, ...5, ...6, ...7]
+   * @returns {[Object,Object,Object,Object,Object,Object,Object]}
```
naxhh: not important but this looks really weird.

This would normally be: `[]Object` or `Array<Object>` or similar
You can clarify that is an array of steps if you feel it's needed

packages/sui-cz/ErrorCommitSaver.js
```diff
@@ -0,0 +1,25 @@
+/* eslint-disable no-console */
+
+const fs = require('fs')
+const path = require('path')
+
+class ErrorCommitSaver {
+  static initErrorListener (commitMessage) {
+    process.on('uncaughtException', function (err) {
+      fs.writeFile(path.join(path.resolve(__dirname), 'tmpCommitMessage.txt'), commitMessage, function () {
```
carlosvillu: Use /tmp for this kind of thing and use a dot file

packages/sui-cz/index.js
```diff
@@ -1,160 +1,13 @@
...
-              })
-            })
-          })
-        } else if (answers.confirmCommit === 'yes') {
-          commit(buildCommit(answers))
-        } else {
-          console.log('Commit has been canceled.')
-        }
-      })
+    ErrorCommitSaver.retrievePreviousCommit(commitString => {
```
naxhh: Why not start always with the normal flow. and only perform the recover if you add a flag? like --recover or similar?

I normally cancel because I've made a mistake, this will make that flow slower every time.

packages/sui-cz/tmpCommitMessage.txt
```diff
@@ -0,0 +1 @@
+feat(global): testing commit message
```
naxhh: guess this wasn't going here?

packages/sui-cz/PrompterManager.js
carlosvillu: in the sui-helper package

packages/sui-cz/PrompterManager.js
carlesnunez: Totally agree 🙃  Sorry about that

packages/sui-cz/tmpCommitMessage.txt
carlosvillu: jejeje ... go to /tmp

packages/sui-cz/index.js
carlesnunez: The flow doesn't be affected in performance terms if you are talking about that. 
On the other hand, this will only save the commit if you reached the last step and you've done a commit. Not in cancel, cmd+c or other cases.

packages/sui-cz/PrompterManager.js
carlesnunez: I'd like to talk about this because I've founded that the usage of callback with fs and exec is a bit anoying for us.

Maybe we could keep this as it is and evolve on a next iteration

packages/sui-cz/package.json
```diff
@@ -12,6 +12,7 @@
   "license": "MIT",
   "devDependencies": {},
   "dependencies": {
+    "colors": "^1.1.2",
```
miduga: Use fixed version instead  with caret.

packages/sui-cz/ErrorCommitSaver.js
```diff
@@ -0,0 +1,25 @@
...
+        console.log(err)
+      })
+    })
+  }
+
+  static retrievePreviousCommit (callback) {
+    fs.readFile(path.join(path.resolve(__dirname), 'tmpCommitMessage.txt'), 'utf8', (err, data) => {
```
davidbarna: Avoid repeating this code

```
path.join(path.resolve(__dirname), 'tmpCommitMessage.txt'),
```

packages/sui-cz/PrompterManager.js
```diff
@@ -0,0 +1,188 @@
...
+          console.log(`${SEP} \n ${buildCommit(answers)} \n ${SEP} \n`)
+          return 'Are you sure you want to proceed with the commit above?'
+        }
+      }
+    ]
+  }
+
+  static startRecoverOldCommitFlow (cz, commit, commitString) {
+    const HR = '\n ******************************************************************\n'
+    console.log(colors.yellow(`${HR} ${commitString} ${HR} -> We founded that you haved problems with the last commit and we saved the message for you 😇.`))
```
davidbarna: typos

`We found that you had problems with the last commit and we saved the message for you`

packages/sui-cz/PrompterManager.js
```diff
@@ -0,0 +1,188 @@
...
+    cz.prompt([
+      {
+        type: 'confirm',
+        name: 'foundedOldCommit',
+        message: '- Would you like to commit with this message?'
+      }
+    ]).then((answers) => {
+      if (answers.foundedOldCommit) {
```
davidbarna: type

Issue #224, opened 2018-02-28
[sui] Use sui-test instead manual karma testing for our own tools


miduga:
We should be using sui-test instead a manual karma file in order to reuse sui-test for testing our own tools.

julian-gernun:
Should we divide this per tool?

miduga:
I've mantained only one sui-test in order to run all tests, so we have the dependency only in one place.

PR #231, opened 2018-03-02
Feature/use sui test on our own package and let use patterns

miduga:
Please review @carlosvillu @carlesnunez @julian-gernun @davidbarna 

This PR fixes https://github.com/SUI-Components/sui/issues/224 🎇

- [x] Let configure pattern and ignorePattern (only browser). 📁
- [x] Use sui-test on our own sui monorepo. 🐶
- [x] Keep it compatible with actual usages. 📦
- [x] Remove not needed anymore packages and 👋 bye configs!


packages/sui-test/bin/sui-test-browser.js
```diff
@@ -10,31 +10,32 @@ const { cleanStack } = require('./karma/util')
...
-        process.stderr.write(
-          chalk.red(cleanStack((err && (err.stack || err.message)) || err)) +
-            '\n'
-        )
-      }
-      process.exit(err.code || 1)
-    })
-})()
+
+const { watch, ci, pattern, ignorePattern } = program
```
carlosvillu: We can use this pattern `/test/**/*Spec.js` like a default pattern to avoid to force the user to use always a pattern flag.

miduga: That's the default pattern. `/test/**/*Spec.js`

https://github.com/SUI-Components/sui/pull/231/files#diff-0326ee9225e945d3c1b4894c5d101c9eR13

But now, default values are in the commander, so the user will be informed with the default path used. :)

PR #232, opened 2018-03-02
feat(sui-mockmock): new sui-mockmock package for testing mocks

davidbarna:
Implementation example here

https://github.schibsted.io/scmspain/frontend-mt--lib-motor/pull/253/files


packages/sui-mockmock/src/index.js
```diff
@@ -0,0 +1,5 @@
+import HttpMocker from './http'
+
+export {
+  HttpMocker
```
carlosvillu: This package preserve the API from the motor`s solution ?!

davidbarna: It doesn't break anything. Actually there's in the desc of this pr a link to the final implementation and everything's fine :+1:

naxhh:
You should document this:

```
"browser": {
    "./test/helpers/mocker/index.js": "./test/helpers/mocker/clientMocker.js",
    "./test/helpers/describers/index.js": "./test/helpers/describers/clientDescribers.js"
  },
```

But I think is much easier to read if we check if we are in a nodejs environment in `src/http/index.js` and choose the mocker from there

carlosvillu:
We must be careful @naxhh to avoid ending with both packages in the bundle. I think that we must do it like you said, but putting `browser: {nock: false}`. Like we can see in the spec

https://github.com/defunctzombie/package-browser-field-spec

davidbarna:
@carlosvillu it's a mocking only for testing purpose. No bundle concern, right ?

carlosvillu:
Yep. But you have to bundle it to test it. And don’t think that you can bundle nock for the browser.

davidbarna:
Done @carlosvillu and @naxhh : https://github.com/SUI-Components/sui/pull/232/commits/a40c2f49394fad8631f8e23e09f4a74dcb7bbb40

packages/sui-mockmock/src/http/index.js
```diff
@@ -0,0 +1,6 @@
+import ServerMocker from './serverMocker'
+import ClientMocker from './clientMocker'
+
+const isNode = Object.prototype.toString.call(typeof process !== 'undefined' ? process : 0) === '[object process]';
```
carlosvillu: That is a good candidate to be at sui-js.

PR #235, opened 2018-03-07
Feat/sui tealium tracker

carlesnunez:
This tool is thinked to be used with the [@s-ui/html-tagger](https://github.com/SUI-Components/sui/tree/master/packages/sui-html-tagger) tool but you can define your own tags and it will work asswell.

### DIFF
--- a/packages/sui-tealium-tracker/README.md
+++ b/packages/sui-tealium-tracker/README.md
@@ -1,0 +1,45 @@
+# @s-ui/html-tagger
+> Tool to tag the HTML of your webpage from a service.
+
+It provides:
+* A way to tag HTML elements, perfect for add tracking on SPAs.
+* Performance focused in order to avoid jank and heavy script load.
+
+# Use:
+
+```js
+import { tagHTML } from '@s-ui/html-tagger'
+// create an object like this
+const tags = {
+  'CSS_SELECTOR_OF_THE_ELEMENT_TO_TAG': {
+    'DATA_TAG_NAME_TO_ADD': 'DATA_TAG_VALUE_TO_ADD'
+  },
+  '.simple-div': {
+    'tracking-tag': 'c_tracking_tag'
+  },
+  '.button': {
+    'tracking-tag': 'c_tracking_button'
+  },
+  '.added-later': {
+    'tracking-tag': 'c_tracking_dynamic'
+  }
+}
+
+// execute the method and use the created tags
+tagHTML({ tags })
+```
+
+# Performance:
+
+* It uses requestIdleCallback if supported, in order to only tag when the browser is idle. If not available, it tries to use requestAnimationFrame in order to tag elements smoothly.
+* It uses MutationObserver if supported in order to only tag elements when these are added.
+
+### Debugger:
+
+You can use this command in the console to check if the tagging has worked correctly.
+
+```js
+console.table(
+  Array.from(document.querySelectorAll('[data-tracking-tag]'))
+)
+```
--- a/packages/sui-tealium-tracker/README.md
+++ b/packages/sui-tealium-tracker/README.md
@@ -1,8 +1,8 @@
 # @s-ui/tealium-tracker
 > Tool to track events like click or change and, if a tag is present on the element dispatch a link event to tealium.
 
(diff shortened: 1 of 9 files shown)


packages/sui-tealium-tracker/src/FunctionThrottler.js
```diff
@@ -0,0 +1,15 @@
+export class FunctionThrottler {
```
davidbarna: Is this behavior different than https://www.npmjs.com/package/just-throttle ?

davidbarna:
I'm not sure if this should be in the sui monorepo.
Putting "every code we share" in sui will make it grow.
I don't know if we should have everything here or we should split it. For instance one mono repo for everything that is dev tooling and another for all libraries, migrating what is in sui-js as separated packages.
Anyway, in fact if we don't make the decision know, doing the split in the future won't be a problem.

packages/sui-tealium-tracker/src/SuiTealiumTracker.js
```diff
@@ -0,0 +1,70 @@
...
+      const option = options[selectedIndex] || {dataset: {}}
+      const {tealiumTag} = option.dataset
+      if (tealiumTag) {
+        this.sendTealiumThrottled(tealiumTag)
+      }
+    })
+  }
+
+  initCustomEventListener () {
+    document.addEventListener(this.customEventName, (evt) => {
```
jcger: You have to remove every listener you are adding in order to avoid memory leaks. At least create a destroy fn

packages/sui-tealium-tracker/src/FunctionThrottler.js
carlesnunez: Don't really know. What this function does is to delay a function call if was previously called before XXX seconds. First execution will always be done instant.

just-throttle does the same?

packages/sui-tealium-tracker/test/SuiTealiumTrackerSpec.js
```diff
@@ -0,0 +1,69 @@
...
+    sendTealiumThrottledSpy = sinon.spy(suiTealiumTracker, 'sendTealiumThrottled')
+  })
+
+  it('should set our customEvent var to test', () => {
+    expect(suiTealiumTracker.customEventName).to.equal('test')
+  })
+
+  it('should call tealium if our button have the tag placed', (done) => {
+    document.getElementById('tealium-click').click()
+    setTimeout(() => {
```
jcger: please avoid setTimeout in unit test, use fake timers as you are already loading sinon

packages/sui-tealium-tracker/package.json
```diff
@@ -0,0 +1,26 @@
...
+    "track",
+    "service",
+    "tealium"
+  ],
+  "author": "Enablers frontend",
+  "license": "ISC",
+  "devDependencies": {
+    "babel-cli": "6.24.1",
+    "babel-preset-sui": "1",
+    "sinon": "^4.4.2"
```
jcger: remove the ^, maybe sinon should be a dependency of sui-test

carlesnunez:
I am agree with what you said. Talking with @carlosvillu and @miduga we decided to put the `sui-tealium-tracker` on a separate repository by the moment and decide what is the main purpose of the `@s-ui/js`

packages/sui-tealium-tracker/README.md
```diff
@@ -0,0 +1,52 @@
+# @s-ui/tealium-tracker
+> Tool to track events like click or change and, if a tag is present on the element dispatch a link event to tealium.
+
+This tool is thinked to be used with the [@s-ui/html-tagger](https://github.com/SUI-Components/sui/tree/master/packages/sui-html-tagger) tool but you can define your own tags and it will work asswell.ls
```
jcger: I think you want to use meant or intended instead of thinked

packages/sui-tealium-tracker/package.json
carlesnunez: I though about this what do you think @miduga could we put some standard dependencies on the sui-test?

carlesnunez:
Closing in order to have an own repo for this package. PR changes will be made in the [new repository](https://github.com/SUI-Components/sui-tealium-tracker)

PR #236, opened 2018-03-07
Sui test e2e

davidbarna:
## Description
New command for e2e tests.
First implementation was done here: https://github.schibsted.io/scmspain/frontend-mt--web-app/pull/233

### DIFF
--- a/packages/sui-test/bin/cypress/cypress.json
+++ b/packages/sui-test/bin/cypress/cypress.json
@@ -1,0 +1,10 @@
+{
+    "fixturesFolder": false,
+    "pluginsFile": false,
+    "supportFile": false,
+    "videoRecording": false,
+    "screenshotOnHeadlessFailure": false,
+    "trashAssetsBeforeHeadlessRuns": true,
+    "viewportWidth": 1240,
+    "viewportHeight": 960
+}
--- a/packages/sui-test/bin/sui-test-e2e.js
+++ b/packages/sui-test/bin/sui-test-e2e.js
@@ -6,8 +6,8 @@ const program = require('commander')
 const { getSpawnPromise, showError } = require('@s-ui/helpers/cli')
 
 const CYPRESS_FOLDER_PATH = path.resolve(__dirname, 'cypress')
-const TESTS_FOLDER = process.cwd() + '/test/e2e'
-const SCREENSHOTS_FOLDER = process.cwd() + '/.tmp/e2e/screenshots'
+const TESTS_FOLDER = process.cwd() + '/test-e2e'
+const SCREENSHOTS_FOLDER = process.cwd() + '/.tmp/test-e2e/screenshots'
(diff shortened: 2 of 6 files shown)


packages/sui-test/bin/cypress/cypress.json
```diff
@@ -0,0 +1,10 @@
+{
+    "fixturesFolder": false,
+    "pluginsFile": false,
+    "supportFile": false,
+    "videoRecording": false,
+    "screenshotOnHeadlessFailure": false,
+    "trashAssetsBeforeHeadlessRuns": true,
+    "viewportWidth": 1240,
```
jcger: what about mobile?

davidbarna: This is just the default. You can change the context in your specs with `cy.viewport`
https://docs.cypress.io/api/commands/viewport.html#Syntax

packages/sui-test/bin/sui-test-e2e.js
```diff
@@ -0,0 +1,51 @@
...
+
+const objectToCommaString = obj => Object.keys(obj)
+  .map(key => `${key}='${obj[key]}'`)
+  .join(',')
+
+program
+  .option('-B, --baseUrl <baseUrl>', 'URL of the site to execute tests (in ./test/e2e/) on.')
+  .option('-S, --screenshotsOnError', 'Take screenshots of page on any failure.')
+  .option('-U, --userAgentAppend <userAgentAppend>', 'Append string to USerAgent header.')
+  .option('-G, --gui', 'Run the tests in GUI mode.')
```
carlosvillu: Where are the watch mode ?!

packages/sui-test/bin/sui-test-e2e.js
```diff
@@ -0,0 +1,51 @@
+#!/usr/bin/env node
+/* eslint no-console:0 */
+
+const path = require('path')
+const program = require('commander')
+const { getSpawnPromise, showError } = require('@s-ui/helpers/cli')
+
+const CYPRESS_FOLDER_PATH = path.resolve(__dirname, 'cypress')
+const TESTS_FOLDER = process.cwd() + '/test/e2e'
```
carlosvillu: We have to avoid run this tests in the others modes

davidbarna: Will do

davidbarna: @carlosvillu and @miduga it's harder that i thought. In sui-test-browser it's fine because it's a glob pattern. But in sui-test-server it's a folder that defaults to `test` : https://github.com/SUI-Components/sui/blob/master/packages/sui-test/bin/sui-test-server.js#L29

I see 2 options:
1. allowing a pattern in my command as it's in server.
2. allowing the ignorePattern option of cypress (it uses minimatch https://docs.cypress.io/guides/references/configuration.html#Global)

In both cases it's a the developer who manages it...
Do you have any better idea ?

carlosvillu: I think that the easier way could be put the e2e folder in the root project. That means make a clear diferencie  between them.

What do you think ?!

davidbarna: Will do !

PR #247, opened 2018-03-28
[sui-bundler] Add compile server config

carlosvillu:
Maybe in the future we could want a command in the CLI. But for now is enough with a simple webpack config to be imported.

### DIFF
--- a/packages/sui-bundler/webpack.config.server.js
+++ b/packages/sui-bundler/webpack.config.server.js
@@ -1,0 +1,31 @@
+const webpackNodeExternals = require('webpack-node-externals')
+const path = require('path')
+
+let webpackConfig = {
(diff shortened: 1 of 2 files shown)


packages/sui-bundler/webpack.config.server.js
```diff
@@ -0,0 +1,31 @@
+const webpackNodeExternals = require('webpack-node-externals')
+const path = require('path')
+
+let webpackConfig = {
+  context: path.resolve(process.cwd(), 'src'),
+  resolve: { extensions: ['*', '.js', '.jsx', '.json'] },
```
davidbarna: Isn't enough with just `'*'` ?

carlosvillu: I think so. 
Jejeje

Issue #138, opened 2017-10-30
[sui-lint] Add prettier config to format our code automatically


miduga:
### Description

We would like to add prettier to our sui-lint in order to format our code automatically based on our linter config rules, so all developers improve development time.

### Additional Information

The implementation should be totally transparent for the user so he shouldn't notice any change but the advantatge of gettign automatically the code formatted before passing the linter, so it gets fixed before even he notice it.

https://github.com/prettier/prettier

davidbarna:
sui-lint will execute a merge of prettier rules and current sui-lint rules 
Add a command `sui-lint format` that executes prettier and `es-lint --fix`

PR #254, opened 2018-04-06
Sui lint prettier

davidbarna:
## Description
* [x] Add a format command
* [x] Make --add-fixes option to add files when modified
* [x] Test if behavior is compliant to our jsx rules. 
* [x] Add --staged flag in sui-precommit rules

Check changes on jsx files: https://github.com/SUI-Components/sui-components/pull/367/files#diff-d65deaf087d1e90de6ccc6d0ba1982b6

## Related Issue
#138

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,27 +3,29 @@
   "version": "0.0.0",
   "private": true,
   "description": "Monorepo for SUI (Schibsted User Interface) packages.",
-  "keywords": [
-    "schibsted",
-    "monorepo",
-    "sui"
-  ],
+  "keywords": ["schibsted", "monorepo", "sui"],
   "scripts": {
-    "phoenix": "rm -Rf node_modules && npm i && sui-mono run-parallel rm -Rf node_modules && sui-mono run-parallel npm i",
+    "phoenix":
+      "rm -Rf node_modules && npm i && sui-mono run-parallel rm -Rf node_modules && sui-mono run-parallel npm i",
     "co": "sui-mono commit",
     "release": "sui-mono release",
     "release:check": "sui-mono check",
     "lint": "sui-lint js && sui-lint sass",
-    "test": "NODE_ENV=test npm run test:client && NODE_ENV=test npm run test:server",
+    "test":
+      "NODE_ENV=test npm run test:client && NODE_ENV=test npm run test:server",
     "test:ci": "npm run test:client -- --ci",
-    "test:client": "NODE_ENV=test sui-test browser -P 'packages/*/test/**/*Spec.js' -I 'packages/*/test/server/*Spec.js'",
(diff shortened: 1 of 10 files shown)


package.json
```diff
@@ -23,7 +23,7 @@
     "wiki:pull": "git subtree pull --prefix=.wiki https://github.com/SUI-Components/sui.wiki.git master --squash",
     "wiki:push": "git subtree push --prefix=.wiki https://github.com/SUI-Components/sui.wiki.git master",
     "commitmsg": "validate-commit-msg",
-    "precommit": "sui-precommit run"
+    "precommit": "sui-lint format-js --staged && sui-precommit run"
```
davidbarna: This is just a proposal :-)
I think we should recommend to do this rather than making `sui-precommit run` to do it automatically...

packages/sui-lint/Readme.md
```diff
@@ -52,6 +52,22 @@ Lints all `**/src/**/*.scss` files in the project, excluding `node_modules`, `li
...
+
+```sh
+$ sui-lint js --staged
+$ sui-lint sass --staged
+$ sui-lint format-js --staged
+```
+
+Same command but applied only on staged files (obtained with `git diff --cached --name-only --diff-filter=d` command).
+
+> **In write mode like with `sui-lint js --fix` or `sui-lint format-js`, the `--staged` option will stage the files again (`git add <file...>`)**
```
davidbarna: We could also to a new option `--add` or `--commit` that will do only this and not make it auto.

package.json
carlosvillu: Why ?!

packages/sui-lint/Readme.md
carlosvillu: I think that can be dangerous add files without my control. For example I usually add file to the stage by chunks. And this feature can add thing to the commit that I really dont want add yet

Maybe with a warning message could be enough

davidbarna: You're right. I'll add a "--add-fixes" flag to do that. Is it ok?

package.json
davidbarna: Because it was adding files to the stage. But I'll change it.

carlosvillu: Yep!

packages/sui-lint/.prettierrc.js
```diff
@@ -0,0 +1,12 @@
+module.exports = {
+  printWidth: 80,
+  tabWidth: 2,
+  singleQuote: true,
+  trailingComma: 'none',
```
AElouai: RIP ;

carlesnunez: maxresdefault

PR #259, opened 2018-04-11
feat(sui-test): create the base of the patcher and add some dependenc…

carlesnunez:
## Description
The next PR have a feature added to the sui-test. Have the purpose of patch the native mocha functions extending its function set and adding a bunch of new functions that will be the responsible of handle with test only for client or server side executions.

### DIFF
--- a/packages/sui-test/README.md
+++ b/packages/sui-test/README.md
@@ -132,6 +132,73 @@ Cypress can be detected as a robot if your server has that kind of protection or
 
 If defined, any error on your tests will create a screenshot of that moment in the `./.tmp/test-e2e/screenshots` folder of your project.
 
+
+# Tools
+
+## Descriptor by environment patcher
+The descriptor by environment is a patch with the purpose of add some extra functionality to our mocha describe and it methods.
+
+### How to import it?
+
+First of all, the patcher MUST BE APPLIED on each test that we want to have the extra methods so at the top of `ourExampleSpec.js` we will add the next code:
+
+```javascript
+import { descriptorsByEnvironmentPatcher } from '@s-ui/test/lib/descriptor-environment-patcher'
+descriptorsByEnvironmentPatcher()
+```
+And that's it, from that line you will have the next methods added to the base of the mocha lib:
+- describe.onlyClient
+- describe.onlyServer
+- describe.onlyClient.only
+- describe.onlyServer.only
+- it.onlyClient
+- it.onlyServer
+- it.onlyClient.only
+- it.onlyServer.only
+### How can I use it?
+Just in the same way as you have been using the describe or it functions earlier:
+
+```javascript
+describe.onlyClient('Users use case', () => {
+  it('should....', () => {
+    // ...
+  })
+})
+
+describe.onlyServer('Users use case', () => {
+  it('should....', () => {
+    // ...
+  })
+})
+```
+
+You can also have `it()` by environment:
+
+```javascript
+describe('Another use case', () => {
+  it.onlyClient('should....', () => {
+    // ...
+  })
+
+  it.onlyServer('should....', () => {
+    // ...
+  })
+})
+```
+
+What about if you want to run only one describe but only for client? You can use the `.only` function in the same way as you've been using earlier.
+
+```javascript
+describe.onlyClient.only('Another use case', () => {
+  it.onlyClient('should....', () => {
+    // ...
+  })
+
+  it.onlyServer('should....', () => {
+    // ...
+  })
+})
+```
 ## Contributing
 
 Please refer to the [main repo contributing info](https://github.com/SUI-Components/sui/blob/master/CONTRIBUTING.md).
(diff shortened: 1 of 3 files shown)


packages/sui-test/README.md
```diff
@@ -132,6 +132,69 @@ Cypress can be detected as a robot if your server has that kind of protection or
...
+### How to import it?
+
+First of all, the patcher MUST BE APPLIED on each test that we want to have the extra methods so at the top of `ourExampleSpec.js` we will add the next code:
+
+```javascript
+import { descriptorsByEnvironmentPatcher } from '@s-ui/test/lib/descriptor-environment-patcher'
+descriptorsByEnvironmentPatcher()
+```
+And that's it, from that line you will have the next methods added to the base of the mocha lib:
+- describe.onlyClient
```
miduga: I've some concerns about the `onlyClient` naming as it collides with the `only` meaning on the mocha world. Maybe a `describe.forClient` could be a better naming?

packages/sui-test/README.md
```diff
@@ -132,6 +132,69 @@ Cypress can be detected as a robot if your server has that kind of protection or
...
+## Descriptor by environment patcher
+The descriptor by environment is a patch with the purpose of add some extra functionality to our mocha describe and it methods.
+
+### How to import it?
+
+First of all, the patcher MUST BE APPLIED on each test that we want to have the extra methods so at the top of `ourExampleSpec.js` we will add the next code:
+
+```javascript
+import { descriptorsByEnvironmentPatcher } from '@s-ui/test/lib/descriptor-environment-patcher'
+descriptorsByEnvironmentPatcher()
```
carlosvillu: That is too verbose.

```
import patcher from '@s-sui/test/lib/patcher'
patcher()
```
Is a little less verbose 

The important here is use it in this case like a default naming. And I know that we are saying to avoid it. But for me in this scenario make sense.

packages/sui-test/src/descriptor-environment-patcher.js
```diff
@@ -0,0 +1,85 @@
...
+const isNode = typeof process === 'object' && process.toString() === '[object process]'
+
+export const descriptorsByEnvironmentPatcher = function descriptorsByEnvironmentPatcher () {
+  /**
+   * This function is the one with the purpose of handle and return the function that should be attached to our only{client | server} patch.
+   * @param {Object} descriptor Is the object that contains the name of the descriptor base function and in the cases that we have nested function calls the nested name too
+   * @param {String} env Can be one of the constants defined in the environments object - {SERVER, CLIENT}
+   * @param {Boolean} isOnly A boolean function to know if we are trying to run a descriptor with the .only Function
+   */
+  function buildFunctionForEnv ({descriptorName, firstLevelFnName} = {}, env, isOnly) {
```
carlosvillu: Don´t merge naming and positional parameters styles.

Use naming parameters over positionals

packages/sui-test/README.md
miduga: I'm not the best naming things. :D Alternative: `describe.runOnClient`. Or if you find another more suitable go ahead. :D

carlesnunez: We have been talking with carlos yesterday about that and we haven't found any alternative so you provide two really good candidates to dismiss the naming confusion. I prefeer describe.runOnClient or describe.runJustOnClient

what do you think?

packages/sui-test/README.md
carlesnunez: As you wish, for me is not a bad thing to have this patcher with a verbose and descriptive name as is something that is modifying the base function set of an external library if is not a really ugly think I would like to remain the import in that way to avoid confusions and questions about 'what is this patcher doing'

davidbarna: To me, even a register like  `import ''@s-sui/test/lib/describe-env-register"` would be enough.

carlesnunez: Yup I've started to do as just a import but the thing is that we need that this logic being executed on EACH import. Why? because each the test describe, it etc return a function that can't be prototype modified :/

I've talked with @carlosvillu about that and it was the best approach that we found.

carlesnunez: If you are really strong with this opinion I don't have any problem to change it but I would like to know and understand the concerns surrounding the default import in this particular case

davidbarna: I get it. Then keep it simple. We could prototype extensions like they do in chai by it'll be ugly.

carlesnunez:
PR Changes done, waiting for check and approval @miduga @carlosvillu @davidbarna

packages/sui-test/README.md
```diff
@@ -147,25 +147,25 @@ import { descriptorsByEnvironmentPatcher } from '@s-ui/test/lib/descriptor-envir
...
-- describe.runOnServer
-- describe.runOnClient.only
-- describe.runOnServer.only
-- it.runOnClient
-- it.runOnServer
-- it.runOnClient.only
-- it.runOnServer.only
+- describe.client
+- describe.server
+- describe.client.only
```
davidbarna: ❤️

PR #261, opened 2018-04-12
Test sui format js

davidbarna:
## Description
Format code with sui-lint format-js

The first commit is the format. The next commits are arrangements to make it work with prettier.


packages/sui-domain/src/NotImplementedUseCase.js
```diff
@@ -23,6 +23,12 @@ export default class NotImplementedUseCase extends UseCase {
    * @override
    */
   execute () {
-    return Promise.resolve(console.warn(`Service ${this._key} is not implemented in the current version of the domain`))
+    return Promise.resolve(
+      console.warn(
+        `Service ${
```
jcger: dislike this!

packages/sui-studio/src/components/navigation/index.js
```diff
@@ -13,90 +13,83 @@ const compliant = getStudioCompliant()
...
+  state = {search: ''}
 
-  _handleChange = (e) => {
-    this.setState({ search: e.target.value })
+  _handleChange = e => {
+    this.setState({search: e.target.value})
   }
 
-  _handleFocus = (e) => {
+  _handleFocus = e => {
```
jcger: IMO this is like the one liner ifs without brackets, it will always be annoying to add a new param

packages/sui-studio/src/components/demo/index.js
```diff
@@ -66,7 +64,7 @@ export default class Demo extends Component {
       playground,
       style,
       themes,
-      ctxt: context,
+      ctxt: context
```
jcger: disagree with this, I think it should end with comma as later on it's easier to update

packages/sui-perf/src/measure.js
```diff
@@ -8,16 +8,21 @@ export const measureFunc = ({mark, stop}) => (label) => (func) => {
...
+export const measureMethod = perf => {
   let measure = measureFunc(perf)
-  return (label) => (obj, methodName) => {
+  return label => (obj, methodName) => {
     var originalMethod = obj[methodName]
     obj[methodName] = function (...args) {
-      const labelText = (typeof label === 'function' && label.call(this, ...args)) || label || methodName
+      const labelText =
+        (typeof label === 'function' && label.call(this, ...args)) ||
+        label ||
```
jcger: IMO it's easier to read if the operator is at the beginning

PR #275, opened 2018-05-10
Windows cmd compatibility

davidbarna:
## Description
This PR was originally created to make sui-test commands compatible with windows.

## Related Issue
`#!/usr/bin/env node` doesn't work on windows.

### DIFF
--- a/packages/sui-helpers/cli.js
+++ b/packages/sui-helpers/cli.js
@@ -1,5 +1,5 @@
 /* eslint no-console:0 */
-const processSpawn = require('cross-spawn')
+const processSpawn = require('child_process').spawn
 const CODE_OK = 0
 const log = console.log
 const colors = require('colors')
@@ -7,6 +7,7 @@ const program = require('commander')
 const execa = require('execa')
 const Listr = require('listr')
 const figures = require('figures')
+const path = require('path')
 const { splitArray } = require('./array')
 
 /**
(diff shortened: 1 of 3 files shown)


packages/sui-helpers/cli.js
```diff
@@ -1,5 +1,5 @@
 /* eslint no-console:0 */
-const processSpawn = require('child_process').spawn
+const processSpawn = require('cross-spawn')
```
davidbarna: the majority of our commands that call commands in sui use sui-helpers. So this should solve a lot of windows related errors.

davidbarna:
@jcger @carlosvillu @miduga @kikoruiz I thought `cross-spawn` was a better solution than mine but after testing on real windows environment it wasn't working. So, I rolled it back and added the solution to parallel commands as well. Please review again !

packages/sui-helpers/cli.js
```diff
@@ -106,7 +107,20 @@ function getSpawnProcess (bin, args, options = {}) {
...
+
+/**
+ * Returns modified command to work on linux, osx and windows
+ * @param  {String} bin     Binary path or alias
+ * @param  {Array} args    Array of args, like ['npm', ['run', 'test']]
+ * @param  {Object} opts to pass to child_process.spawn call
+ * @returns {Object} {bin, args, options}
+ */
+function getArrangedCommand (bin, args, opts) {
+  return path.isAbsolute(bin) // check if it's a file or an alias
```
jcger: I think you could explain a bit better with code/comments why file or alias is checked with `isAbsolute` and why that's enough to know that it is being run on a windows/unix env

davidbarna: Done !

PR #277, opened 2018-05-22
Docs/add readme for hash

miduga:
- [x] Sort README in order to be alphabetically sorted the libraries.
- [x] Add info about the createHash on the README.


miduga:
Yes @davecarter. In this case, we use md5 for NON-SECURE things, like hashing of files naming or cache keys but not for user passwords so on. :) The good thing about md5 is that is small, very fast and it has few collisions yet no secure.

PR #298, opened 2018-06-13
fix(sui-bundler): sui-bundler dev not using babel-preset-sui

davidbarna:
## Description
dev config had a bad format. So babel-preset-sui was not used and "env" was used instead.
This causes problems in studios that use spread operators in they demo files for instance.

### DIFF
--- a/packages/sui-bundler/shared/module-rules-babel.js
+++ b/packages/sui-bundler/shared/module-rules-babel.js
@@ -1,0 +1,9 @@
+module.exports = {
+  test: /\.jsx?$/,
+  exclude: /node_modules(?!\/@s-ui\/studio\/src)/,
+  loader: 'babel-loader',
+  query: {
+    babelrc: false,
+    presets: ['sui']
+  }
+}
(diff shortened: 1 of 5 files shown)


packages/sui-bundler/webpack.config.dev.js
```diff
@@ -44,18 +45,7 @@ let webpackConfig = {
   ],
   module: {
     rules: [
-      {
-        test: /\.jsx?$/,
-        include: /src/,
```
jcger: isn't this include needed?

packages/sui-bundler/shared/module-rules-babel.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  test: /\.jsx?$/,
+  exclude: /node_modules(?!\/@s-ui\/studio\/src)/,
+  loader: 'babel-loader',
+  query: {
+    babelrc: false,
```
davidbarna: This was only in dev but i think it must be applied to all

davidbarna: Our problem was in sui-studio start

packages/sui-bundler/webpack.config.dev.js
davidbarna: It works. And checkout the rest of the configs. No trace of it.
Will recheck anyways.

PR #307, opened 2018-06-20
fix(sui-bundler): fix bundling in windows

davidbarna:


packages/sui-bundler/shared/define.js
```diff
@@ -1,14 +1,9 @@
...
-let pwd = process.cwd()
-
 // hack for Windows, as process.env.PWD is undefined in that environment
 // https://github.com/mrblueblue/gettext-loader/issues/18
-// Moreover, to have same caseing we need to map it.
 if (process.platform === 'win32') {
-  const [driveLetter, path] = pwd.split(':')
-  pwd = [driveLetter.toLowerCase(), path].join(':')
-  process.env.PWD = pwd
+  process.env.PWD = process.cwd()
```
davidbarna: Without linking process.cwd() still returns "C:", but the rest of the paths are now in "C:" too. I supposed it has to do sth with linking...

packages/sui-bundler/shared/module-rules-babel.js
```diff
@@ -1,6 +1,8 @@
+const {sep} = require('path')
+
 module.exports = {
   test: /\.jsx?$/,
-  exclude: /node_modules(?!\/@s-ui\/studio\/src)/,
+  exclude: new RegExp(`node_modules(?!${sep}@s-ui${sep}studio${sep}src)`),
```
davidbarna: This change had been un done but ends up being necessary. Probably sth that has to do with linking.

PR #312, opened 2018-06-26
fix(sui-domain): fix: anemic models toJSON not recursive

davidbarna:

### DIFF
--- a/packages/sui-domain/src/AnemicModel.js
+++ b/packages/sui-domain/src/AnemicModel.js
@@ -2,13 +2,13 @@
  * Recursively calls toJSON on each member
  * @param {Object} obj
  */
-const toJSON = obj => {
-  Object.keys(obj).forEach(key => {
-    obj[key] = obj[key] instanceof AnemicModel ? obj[key].toJSON() : obj[key]
-    obj[key] = obj[key] instanceof Object ? toJSON(obj[key]) : obj[key]
-  })
-  return obj
-}
+const mapValuesToPlainObjects = obj =>
+  Object.keys(obj).reduce((o, key) => {
+    o[key] = obj[key]
+    if (o[key] instanceof AnemicModel) o[key] = o[key].toJSON()
+    if (o[key] instanceof Object) o[key] = mapValuesToPlainObjects(o[key])
+    return o
+  }, obj instanceof Array ? [] : {})
 
 /**
  * Converts anemic instance to an object with public props
@@ -50,6 +50,6 @@ export default class AnemicModel {
    * @returns {*}
    */
   toJSON() {
-    return toJSON(anemicInstanceToObject(this))
+    return mapValuesToPlainObjects(anemicInstanceToObject(this))
   }
 }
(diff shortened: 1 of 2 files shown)


packages/sui-domain/src/AnemicModel.js
```diff
@@ -1,3 +1,28 @@
...
+  return obj
+}
+
+/**
+ * Converts anemic instance to an object with public props
+ * @param {AnemicModel} obj
+ */
+const anemicInstanceToObject = obj =>
+  Object.keys(obj)
+    .map(key => key.replace('_', ''))
```
davidbarna: I'm a bit concern to loop more times but this code is cleaner this way.
I can change it to a unique loop

packages/sui-domain/src/AnemicModel.js
```diff
@@ -1,3 +1,28 @@
+/**
+ * Recursively calls toJSON on each member
+ * @param {Object} obj
+ */
+const toJSON = obj => {
```
carlesnunez: The naming of toJSON could lead us to missunderstand what are we calling. We could think that we are calling `obj.toJSON()` instead that call a local toJSON function.

packages/sui-domain/src/AnemicModel.js
```diff
@@ -1,3 +1,28 @@
...
+  Object.keys(obj).forEach(key => {
+    obj[key] = obj[key] instanceof AnemicModel ? obj[key].toJSON() : obj[key]
```
carlesnunez: For me is not so clean reassign a variable to itself just in order to have it as a ternary operation instead than on an if. I'll change it by this
```javascript
obj[key] = obj[key] instanceof AnemicModel ? obj[key].toJSON() : obj[key] instanceof Object ? toJSON(obj[key]) : obj[key]
```

Or if it looks unreadable (I assume that yes)

```javascript
if (obj[key] instanceof AnemicModel) {
    obj[key] = obj[key].toJSON() 
} else if (obj[key] instanceof Object) {
    obj[key] = toJSON(obj[key])
}

Just in order to increase the understability of the code.
```

packages/sui-domain/src/AnemicModel.js
carlesnunez: Yup it iterates more but performance here doesn't seems to be a problem.... 👌

packages/sui-domain/src/AnemicModel.js
davidbarna: I agree, but then I would need:
```js
if (obj[key] instanceof AnemicModel) {
    obj[key] = toJSON(obj[key].toJSON())
} else if (obj[key] instanceof Object) {
    obj[key] = toJSON(obj[key])
}
```

Agree ?

packages/sui-domain/src/AnemicModel.js
davidbarna: I'm assuming here that the obj was created brand new because it's the same scope (which is the case).
But I agree I could be more functional. Will change it. Thanks.

packages/sui-domain/src/AnemicModel.js
carlesnunez: I don't get the point of the toJSON naming function but it looks more clearer than the older way! 👍 🔝

Issue #299, opened 2018-06-13
[sui-ssr && sui-react-initial-props] Always re-rendering once navigate between routes.


carlesnunez:
It is more a problem with loadPage than with sui-ssr itself but assuming that loadPage is strictly attached to an ssr environment I associated it with ssr.
ssr@latest
react-initial-props@latest

### Description

react-router **is re-rendering its pages always** that we navigate due to a change-ref problem.

I realised about the problem when I've tried to navigate from one route to a nested one. 

For example let's imagine that we have this route scenario:

```javascript
  <Router>
    <Route path="/creditos/auto-renueva" getComponent={LoadAutoRenewPage}>
      <Route path="alertas" getComponent={LoadAlertsPage} />
    </Route>
  </Router>
```

And the next interaction scenario:
1. Lets say the user enters the app at `/creditos/auto-renueva`.

    | Component | Lifecycle Hooks *expected* | Lifecycle Hooks *called*
    |-----------|------------------------| ------------------------|
    | LoadAutoRenewPage | (1) `componentDidMount` | (1) `componentDidMount` |
    | LoadAlertsPage | N/A | N/A |

2. Now they navigate from `/creditos/auto-renueva` to `/creditos/auto-renueva/alertas`

    | Component | Lifecycle Hooks *expected* | Lifecycle Hooks *called*
    |-----------|------------------------| ------------------------|
    | LoadAutoRenewPage | (1) `componentWillReceiveProps `, (3) `componentDidUpdate` | (2) `componentDidMount` |
    | LoadAlertsPage | (2) `componentDidMount` | (1) `componentDidMount`

**What should happen**
- `LoadAutoRenewPage` should get `componentWillReceiveProps` and `componentDidUpdate` because it
    stayed rendered but just received new props from the router (like:
    `children`, `params`, `location`, etc.)
- `LoadAlertsPage` is mounted for the first time.

**What's happening**
- `LoadAutoRenewPage` is mounted for the ~first time~
- `LoadAlertsPage` is mounted for the first time.

More info on react-router v3 component mounting lifecycle [here](https://github.com/ReactTraining/react-router/blob/v3/docs/guides/ComponentLifecycle.md)

### Why is this happening?

Looking through the code I started trying to just do an async import without any usage of loadPages function.

Changing this:

```javascript
const LoadAutoRenewPage = loadPage(contextFactory, () =>
  import(/* webpackChunkName: "AutoRenueva" */ './pages/credits/AutoRenew')
)
[...]

PR #315, opened 2018-06-27
fix(sui-react-initial-props): fix unnecesary unmount of pages

miduga:
## Description
It seems we're unmounting and mounting without need some pages. This is a first approach in order to address the problem. We need further testing in order to check this is not breaking anything, if this change let us to a major and how this change impact actual pages.

## Related Issue
https://github.com/SUI-Components/sui/issues/299

## Example
More info in the issue.

### DIFF
--- a/packages/sui-react-initial-props/src/loadPage.js
+++ b/packages/sui-react-initial-props/src/loadPage.js
@@ -3,8 +3,18 @@ import PropTypes from 'prop-types'
 
 import withInitialProps from './withInitialProps'
 import createClientContextFactoryParams from './createClientContextFactoryParams'
+// object to store pages already created with the needed context
+let pagesMemoization = {}
 
 const EMPTY_GET_INITIAL_PROPS = () => Promise.resolve({})
+
+const createPageOnClientWithContext = ({Page, contextFactory, routeInfo}) => {
+  return contextFactory(createClientContextFactoryParams()).then(context => {
+    const PageWithContext = withInitialProps({context, routeInfo})(Page)
+    pagesMemoization[Page.name] = PageWithContext
+    return PageWithContext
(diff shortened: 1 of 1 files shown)


carlesnunez:
Going to test it on real test environment

davidbarna:
Build is failing but it might have nothing to do with your code. Just merge once your done even if the check fails.

carlesnunez:
Seems that keep's not working. The current behavior with the fix is:

-> First render (from server) `componentDidMount`
-> Second render (from frontend, after a redirection) `componentDidMount` (BAD BEHAVIOR)
-> Next renders `componentDidUpdate` (GOOD BEHAVIOR from here)

As I told to @miduga this is happening because the first render is rendering a component with a different ref than the second render. This is happening due to we only fixed the case from client, not from the SSR.

carlesnunez:
Whats the state of this pr?

miduga:
Let's get tomorrow a momento in order to sit together and try new approaches @carlesnunez . :)

carlesnunez:
Now that we've returned form holidays we can catchup with that when you want :)

miduga:
We have discussed this some times in order to get in our JIRA board but always there is some other more important task to do. If this is causing a real problem on any implementation, then we definitely should focus on it, talk with our product owner and prioritize it accordingly. I understand that conceptually speaking is not correct but right now this is being used as a feature on all the products for getting the initial props for the component and perform some tracking and stuff in componentDidMount. We will arrange a meeting in order to understand better if this is causing a major problem in your project and thus prioritize it if needed or closing the ticket for now.

carlesnunez:
Thank you for considering this in order to decide if is a problem or not! 👍

PR #320, opened 2018-07-05
sui-deploy default for docker deployments

davidbarna:

### DIFF
--- a/packages/sui-deploy/README.md
+++ b/packages/sui-deploy/README.md
@@ -20,8 +20,24 @@ When installed, a new CLI `sui-deploy` (node_modules/.bin/sui-deploy) is automat
 Deploy a folder as an static SPA:
 
 ```sh
+$ sui-deploy <name-of-project> [build-folder=./public] --now
 $ sui-deploy spa <name-of-project> [build-folder=./public] --now
 ```
+
+### `sui-deploy <name> [folder]`
+
+Deploys given folder.
+Handles:
+* `docker run` if DockerFile is found.
+* `npm run start` if package.json is found.
+* `serve .` otherwise.
+
+
+### `sui-deploy spa <name> [folder]`
+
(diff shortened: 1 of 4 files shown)


packages/sui-deploy/README.md
```diff
@@ -35,7 +51,7 @@ $ sui-deploy spa 'test-project' --now -p
 
 Deploys to test-project.now.sh
 
-#### `-a, --auth <user:password>`
+#### `-a, --auth <user:password>` (only for spa)
```
carlosvillu: I dont know how or if we have to say something here, but you can do the same with the sui-ssr.

carlosvillu: Can you put an example of use sui-deployer with sui-ssr ?!

packages/sui-deploy/README.md
```diff
@@ -20,9 +20,25 @@ When installed, a new CLI `sui-deploy` (node_modules/.bin/sui-deploy) is automat
 Deploy a folder as an static SPA:
 
 ```sh
+$ sui-deploy <name-of-project> [build-folder=./public] --now
```
carlosvillu: maybe `sui-deploy folder ...` make more sense, ist ?!

packages/sui-deploy/README.md
davidbarna: Can can't do auth with now without actually modifying the package.json if you use serve. In case of sui-ssr we would need to implement it in server.js. I don't want to couple sui-deploy to any file structure or code. And @carlesnunez already implement the auth thing in sui-ssr.
We are deploying a folder no matter who created it.
Here you have an example: https://github.schibsted.io/scmspain/frontend-mt--web-app/pull/295/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R36

PR #321, opened 2018-07-06
Feature/mmaa 4032 add auth to ssr

carlesnunez:
## Description
In order to allow developers to have a ssr server protected over user and password I've added the --auth parameter to the ssr.

The parameter is injecting two ENV vars on the dockerfile that will be getted from express on request run-time and passed to a middleware that will prompt us asking for credentials.

I've also improved the console output in order to be more user friendly.

## Example
ezgif com-video-to-gif 9

### DIFF
--- a/packages/sui-ssr/archive/index.js
+++ b/packages/sui-ssr/archive/index.js
@@ -1,15 +1,22 @@
 const path = require('path')
 const fs = require('fs')
 const archiver = require('archiver')
-
+const program = require('commander')
+const authDefinitionBuilder = require('./authDefinitionBuilder')
 module.exports = ({outputZipPath, pkg}) =>
   new Promise((resolve, reject) => {
+    let authVariableDefinition = program.auth
+      ? authDefinitionBuilder(program.auth.split(':'))
+      : ''
     const output = fs.createWriteStream(outputZipPath)
     const archive = archiver('zip', {
       zlib: {level: 9}
     })
 
-    output.on('close', resolve)
+    output.on('close', () => {
+      console.log(' -> Success ✅'.green)
+      resolve()
+    })
     archive.on('error', reject)
     archive.pipe(output)
     archive.append(
@@ -18,7 +25,13 @@ module.exports = ({outputZipPath, pkg}) =>
         .replace('{{name}}', pkg.name),
       {name: 'pm2.json'}
     )
-    archive.file(path.join(__dirname, 'Dockerfile'), {name: 'Dockerfile'})
+
+    archive.append(
+      fs
+        .readFileSync(path.join(__dirname, 'Dockerfile.tpl'), 'utf8')
+        .replace('{{AUTH_VARIABLES}}', authVariableDefinition),
+      {name: 'Dockerfile'}
+    )
     archive.directory(path.join(process.cwd(), 'public'), 'public')
     archive.directory(path.join(process.cwd(), 'statics'), 'statics')
     archive.directory(path.join(process.cwd(), 'server'), 'server')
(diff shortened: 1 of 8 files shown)


packages/sui-ssr/package.json
```diff
@@ -17,10 +17,11 @@
   },
   "dependencies": {
     "@s-ui/hoc": "1.5.0",
-    "archiver": "2.1.1",
+    "archiver": "^2.1.1",
     "commander": "2.15.1",
     "copy-paste": "1.3.0",
     "express": "4.16.3",
+    "express-basic-auth": "^1.1.5",
```
miduga: Exact version here.

packages/sui-ssr/bin/sui-ssr-archive.js
```diff
@@ -3,14 +3,17 @@
...
 const rimraf = require('rimraf')
 const path = require('path')
-
 const archive = require('../archive')
+require('colors')
 
-const OUTPUT_ZIP_PATH = path.join(process.cwd(), 'server.zip')
 const pkg = require(path.join(process.cwd(), 'package.json'))
-
+const REMOVE_ZIP_PATH = path.join(process.cwd(), '*.zip')
```
carlosvillu: too risky this command!

carlesnunez: Assuming that the path will be our webapp folder and assuming also that we will have just one zip at time.... will it throw problems?


Don't really know an alternative to that assuming that if we want to add on a next iteration a name param to the command we will have to 'save' the last name in order to remove the 'last' .zip builded.


What will you do in order to change it?

packages/sui-ssr/package.json
```diff
@@ -17,10 +17,11 @@
   },
   "dependencies": {
     "@s-ui/hoc": "1.5.0",
-    "archiver": "2.1.1",
+    "archiver": "^2.1.1",
```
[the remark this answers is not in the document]
carlesnunez: done!

packages/sui-ssr/package.json
carlesnunez: done!

packages/sui-ssr/archive/index.js
```diff
@@ -1,15 +1,22 @@
...
+      ? authDefinitionBuilder(program.auth.split(':'))
+      : ''
     const output = fs.createWriteStream(outputZipPath)
     const archive = archiver('zip', {
       zlib: {level: 9}
     })
 
-    output.on('close', resolve)
+    output.on('close', () => {
+      console.log(' -> Success ✅'.green)
```
ddmarin94: Maybe, could be interesting have the size of the file in bytes, like this --> 

```javascript
console.log(archive.pointer() + ' total bytes');
```
?????

carlesnunez: sure! I'll do that and if someone thing that doesn't proceed I'll delete it.

packages/sui-ssr/bin/sui-ssr-archive.js
carlosvillu: If you will use a glob expresión to delete files from the user workspace. At least show and confirm which files you’ll end up deleted.

carlesnunez: Another thing that we could do is to create all files with a sufix like <name>-build.zip

carlosvillu: +1 for *-sui-ssr.zip prefix.

carlesnunez:
PR changes done!

PR #322, opened 2018-07-06
Feature/mmaa 4032 add custom name to archive

carlesnunez:
## Description
The next PR ads a feature to sui-ssr that allow us to:

- Pipe our compressed folder to stdout stream
- Select a custom name over --outputFileName command.

## Example
More info on readme and gif:
ezgif com-video-to-gif 10

### DIFF
--- a/packages/sui-ssr/README.md
+++ b/packages/sui-ssr/README.md
@@ -1,9 +1,11 @@
 # sui-ssr
+
 > Plug SSR to you SUI SPA.
 
 SSR can be tought to configure and maintain. SSR handles that for you providing:
-* SSRaS Server-Side Rendering as a Service
-* Server improvements shared accross projects
+
+- SSRaS Server-Side Rendering as a Service
+- Server improvements shared accross projects
 
 ## Installation
 
@@ -16,7 +18,6 @@ npm install @s-ui/ssr --save
 Generate a static version of the server w/out dependencies in the server folder.
 
 ```
-
   Usage: sui-ssr-build [options]
 
   Options:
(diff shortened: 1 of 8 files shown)


packages/sui-ssr/README.md
```diff
@@ -47,11 +48,26 @@ It will, over parameter, make that the express server run over a username and pa
...
+    $ sui-ssr archive --outputFileName=myFile // output: myFile-sui-ssr.zip
+```
+
+### IMPORTANT!
+
+If no outputFileName is provided it will pipe the standard output stream `process.stdout`
+
+## User the ssr output as stream
+
+It use the stdout strema so you can do things like:
```
davidbarna: typo stream

packages/sui-ssr/README.md
```diff
@@ -47,11 +48,26 @@ It will, over parameter, make that the express server run over a username and pa
     -C, --clean  Remove previous zip
     -h, --help   output usage information
     -A, --auth <username:password> Will build the express definition under authentication htpassword like.
+    -N, --outputFileName <outputFileName> A string that will be used to set the name of the output filename. Keep in mind that the outputFilename will have the next suffix <outputFileName>-sui-ssr.zip
```
davidbarna: To me, adding the suffix makes no sense.

davidbarna: usually `-o` is used for output in many CLIs.

davidbarna:
You are breaking the API without major release. I know sui-ssr is not supposed to be used yet. But, I have a PR using it and moreover we're not sure if someone is using it in a personal project for instance. Remember to apply OS mindset.

packages/sui-ssr/README.md
carlesnunez: The sufix was made in order to have 'tracked' which were the files that we created using the tool to delete on -C flag selected.

carlesnunez:
What should I do regarding the version thing? Increase a major? Don't do anything but keep in mind retrocompat for next releases? @davidbarna @carlosvillu

carlesnunez:
I'll update a major in order to non be api breaker 😄

Issue #152, opened 2017-11-15
[sui-mono] Release single package


jcger:
We want to be able to release a single package when running `sui-momo release` in order to avoid releasing  packages from others

PR #328, opened 2018-07-10
Feature/sui mono single package release

jcger:
#152

### DIFF
--- a/packages/sui-mono/README.md
+++ b/packages/sui-mono/README.md
@@ -1,4 +1,5 @@
 # sui-mono
+
 > Simple CLI for monorepo/multipackage.
 
 Sui-mono is a tool that aims to simplify managment for monorepo/mutlipackage projects but it also works with monopackage projects.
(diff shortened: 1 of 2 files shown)


davidbarna:
Good job. FYI I've been thinking a way to apply the package filter to any of sui-mono commands with a common option like so:

```sh
sui-mono release --pkg="category/*"
sui-mono commit-all -t=fix -m="A commit" --pkg="category/package"
sui-mono phoenix --pkg="category/*"
```

I'm just saying in case you wanna make your change more prepared for a common option for all commands.

Plus *the concept of category and component is a sui-studio concept, not sui-mono*.

jcger:
I wasn't aware that there is an option called "deepLevel". Maybe this PR is useless as it doesn't includes deepLevel > 2, or maybe sui-mono does to much and should not allow deepLevel > 2

davidbarna:
@jcger you may have to consider to close it or refactor 😢

jcger:
I'll handle it as one and call it scope, so it would be possible to release a/very/long/but/single/deep/level/scope

PR #334, opened 2018-07-13
fix(sui-bundler): fix Travis CI broken when used node 10

miduga:
Fixes error build problems with Node 10.

image


davidbarna:
Is node 10 allowed ? Are we using it somewhere already ?

miduga:
Yes @davidbarna, personal projects. 😋I would say we're not officially supporting node 10 but making node 10 compatible is "free".

PR #340, opened 2018-07-31
adds number formating functions

davecarter:
This small utility has the purpose of being a centralized tool to format numbers in thousands with dots or commas.

## Description
I'm not using `toLocaleString` to avoid incompatibility issues with browsers like IE or Safari Mobile.

## Usage
```javascript
const numToFormat = 100000
numberWithDots({ value: numToFormat })  // result : '100.000'
```
Please review @SUI-Components/developers

### DIFF
--- a/packages/sui-js/src/number/index.js
+++ b/packages/sui-js/src/number/index.js
@@ -1,0 +1,7 @@
+const numberWithDots = ({ value = 0 }) =>
+  `${value.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ".")}`;
+
+const numberWithCommas = ({ value = 0 }) =>
+  `${value.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",")}`;
+
+export { numberWithDots, numberWithCommas };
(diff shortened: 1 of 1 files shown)


packages/sui-js/src/number/index.js
```diff
@@ -0,0 +1,7 @@
+const numberWithDots = ({ value = 0 }) =>
+  `${value.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ".")}`;
+
+const numberWithCommas = ({ value = 0 }) =>
+  `${value.toString().replace(/\B(?=(\d{3})+(?!\d))/g, ",")}`;
```
jcger: you are repeating code, what about a factory?

```javascript
const formatFactory = ({delimeter}) => ({value = 0}) =>
  `${value.toString().replace(/\B(?=(\d{3})+(?!\d))/g, delimeter)}`;

const numberWithCommas = formatFactory({delimeter: ','})
```

davecarter: that's cool. Thanks :)

PR #346, opened 2018-08-10
Err404 thrown when parameter dont have same order readme update

tureey:
Updated the `README.md` for noticing an error that someone else can find with the usage of the `mockmock package`

## Description
The `mockmock package` needs to have the same parameter order in the mocked and the original http request


tureey:
Is there any reason why the ![correct] and ![incorrect] are not seen properly at least in the preview?

PR #361, opened 2018-08-23
Feature/sui ssr  add hooks

carlosvillu:


packages/sui-ssr/README.md
```diff
@@ -70,6 +70,29 @@ It use the stdout stream so you can do things like:
...
+
+export default {
+  [TYPES.LOGGING]: (req, res, next) => {
+    console.log(req.url)
+    next()
+  }
+}
+```
+
+Here we are define a logger to the console directly. Each hook type could be or a middleware function or an array of middlewares functions.
```
jcger: we implement instead of we are define, I would even use something like: here we implement a direct log into the console or something like that.

packages/sui-ssr/README.md
```diff
@@ -70,6 +70,29 @@ It use the stdout stream so you can do things like:
...
+
+You can checks with hooks there is avalible in the hooks-types.js file.
```
jcger: hooks are available instead of hooks there is avalible

packages/sui-ssr/README.md
```diff
@@ -70,6 +70,29 @@ It use the stdout stream so you can do things like:
...
+
+There is two default hooks for 404 and 500 errors. both will look for a 404.html or 500.html file in the src folder and show this file. If you dont define this files, you will get an generic error page.
```
jcger: there are instead of there is
get a generic  error instead of get an generic

packages/sui-ssr/compiler/server.js
```diff
@@ -13,6 +14,14 @@ module.exports = ({outputPath}) =>
       filename: 'index.js'
     },
     externals: undefined,
+    plugins: [
+      ...(serverConfig.plugins || []),
+      new CopyWebpackPlugin([
+        {from: '404.html', to: '../public'},
+        {from: '500.html', to: '../public'},
+        {from: '200.html', to: '../public'}
```
jcger: what is the 200.html file?

carlosvillu: I have to removed!

PR #363, opened 2018-08-24
Feat/adds ur ifriendly string normalizer

davecarter:
## Description
This function normalices capital letters, spaces, accents and any other non standard ASCII character into an URI friendly string.

## Example
`console.log(uriNormalizer("España y L'Hóspitalët")) // espana-y-l-hospitalet`

Please review

### DIFF
--- a/packages/sui-js/src/string/uri-normalizer.js
+++ b/packages/sui-js/src/string/uri-normalizer.js
@@ -1,7 +1,9 @@
+import slugify from 'slugify'
+
+const customCharMap = {
+  "'": '_'
+}
 export const uriNormalizer = str => {
-  return str
-    .toLowerCase()
-    .replace(/[\s']/g, '-')
-    .normalize('NFD')
-    .replace(/[\u0300-\u036f]/g, '')
+  slugify.extend(customCharMap)
+  return slugify(str).toLowerCase()
 }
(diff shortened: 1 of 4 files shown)


packages/sui-js/src/string/uri-normalizer.js
```diff
@@ -0,0 +1,7 @@
+export const uriNormalizer = str => {
+  return str
+    .toLowerCase()
+    .replace(/[\s']/g, '-')
+    .normalize('NFD')
```
davidbarna: normalize is not supported in IE and AFAIK it's not in sui-polyfills.

packages/sui-js/src/string/uri-normalizer.js
```diff
@@ -0,0 +1,7 @@
+export const uriNormalizer = str => {
```
davidbarna: Have you considered using one of the uri normalizers already in NPM ?

miduga:
In case you need a npm package, you could use this one:
https://www.npmjs.com/package/slugify

davecarter:
We have decided to use `i18n` `url` localization features instead. So I'm closing without merge.

carlesnunez:
We are using it at milanuncios too :)

PR #367, opened 2018-08-29
Feature/upgrade to babel 7

miduga:
Upgrade to babel@7.

- [x] Tested on Real Estate SPA and Server.
- [ ] Test on Sui-Studios.
- [ ] Test on another web.
- [ ] Test on sui-ssr.


packages/babel-preset-sui/lib/index.js
```diff
@@ -15,33 +44,18 @@ module.exports = {
...
-      require('babel-plugin-transform-react-remove-prop-types').default,
-      {
-        mode: 'wrap'
-      }
-    ]
-  ],
-  env: {
-    development: {
-      plugins: cleanList([
-        isInstalled(['preact', 'react'], 'react-hot-loader/babel')
```
carlosvillu: We need this for the hot reloading: https://github.com/gaearon/react-hot-loader#getting-started

packages/babel-preset-sui/lib/index.js
```diff
@@ -1,12 +1,41 @@
-const isInstalled = require('./is-installed')
-const cleanList = require('./clean-list')
+console.log('new babel')
 
-module.exports = {
-  presets: cleanList([
+function plugins(api, opts) {
```
carlosvillu: Are we crazy enough ?! 
https://babeljs.io/docs/en/babel-plugin-proposal-optional-chaining#via-babelrc-recommended

packages/babel-preset-sui/lib/index.js
```diff
@@ -1,12 +1,41 @@
...
+      require('@babel/preset-env').default,
       {
         debug: false,
+        ignoreBrowserslistConfig: true,
+        loose: true,
+        modules: false,
+        // Exclude transforms that make all code slower
+        exclude: ['transform-typeof-symbol'],
         targets: {
           node: '6.0.0',
```
carlosvillu: Maybe we can set up this at Node 8

packages/babel-preset-sui/lib/index.js
```diff
@@ -15,33 +44,18 @@ module.exports = {
...
-      plugins: cleanList([
-        isInstalled(['preact', 'react'], 'react-hot-loader/babel')
-      ])
-    }
-  }
+    [require('@babel/preset-react').default, {useBuiltIns: true}]
+  ]
 }
+
+module.exports = (api, opts) => ({
```
carlosvillu: Could we use this new API to do what we were talking the last Sunday?

packages/babel-preset-sui/lib/index.js
midudev: hahaha let's do it as a new feature as soon as we have babel7. :)

packages/babel-preset-sui/lib/index.js
midudev: Yikes! I've removed it for testing, I've added a pending task in order to add it. Good catch!

packages/babel-preset-sui/lib/index.js
midudev: I wanted to keep compatibility for now, to avoid doing this step forward in the same PR. Also, FC is still using Node 6 in pro.

PR #373, opened 2018-08-31
feat(sui-i18n): this PR ads a dot to the regex of the sui-i18n

carlesnunez:
## Description
Modified the regex in order to admit dots on our urls. So we can have 'fake' urls pointing to .php's .html's etc....

### DIFF
--- a/packages/sui-i18n/src/slugify.js
+++ b/packages/sui-i18n/src/slugify.js
@@ -11,7 +11,7 @@ export function slugify(str) {
   return str
     .toLowerCase()
     .replace(/.{1}/g, replaceCharIfNeeded)
-    .replace(/[^a-z0-9 -]/g, '') // remove invalid chars
+    .replace(/[^a-z0-9 - .]/g, '') // remove invalid chars only letters numbers, dashes and dots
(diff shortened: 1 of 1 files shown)


packages/sui-i18n/src/slugify.js
```diff
@@ -11,7 +11,7 @@ export function slugify(str) {
   return str
     .toLowerCase()
     .replace(/.{1}/g, replaceCharIfNeeded)
-    .replace(/[^a-z0-9 -]/g, '') // remove invalid chars
+    .replace(/[^a-z0-9 - .]/g, '') // remove invalid chars only letters numbers, dashes and dots
```
davidbarna: What about not parching and using what @davecarter has just done [here](https://github.com/SUI-Components/sui/pull/363/files) ?
Our slugs will indeed be more consistent.

carlesnunez: I wont want to change the tool drastically so if we do that I'll increase the sui i18n one major in order to be pretty sure to avoid side effects

carlesnunez: @davidbarna @davecarter decided to close his tool so... this is the only place with an slufigy

PR #376, opened 2018-09-03
Load env vars from file on runtime, buid app config object and pass it to the SPA

betobe:
## Description
The main purpose is about being able to set environmental variables on runtime so that:
- you don't have to build the app one time per environment.
- It means that this ease, simplifies and speeds up your CI/CD process.
- Also the code and config are separated so that your app becomes environment agnostic.
- And as your app is built just once you can trust that what you test is exactly what you deploy.

## Additional info
- Note that env file was named with "public" word so that developers have a clue to not add secrets in it. If a secret was needed, the same approach could be applied in an implementation where secrets are read from a `private-env.yml` file with the exception that this data should not be passed to the SPA in any case.
- req.hostname is added to appConfig by default. In the NC project the hostname is needed in order to identify the site the user is going to see and it is used to construct the domain layer. For any other case I thought it could be also useful to have the hostname set from the server by default.

## Example
Create a yaml file and place it in the root directory of your SPA.
```
# mySpa/public-env.yml
API_SOME_ENDPOINT: test.some.com/some/service
API_SOME_OTHER_ENDPOINT: test.some.com/some-other/service
```

Now you can have access to those envs from your app's context factory. You can use it to construct your data layer config.
```
# mySpa/src/contextFactory.js
export default function contextFactory({
  appConfig = {},
  cookies,
  isClient,
  pathName,
  req = {},
  userAgent
}) {
  const domain = new MyDomain(appConfig.envs)
...
```

### DIFF
--- a/packages/sui-ssr/README.md
+++ b/packages/sui-ssr/README.md
@@ -62,9 +62,9 @@ It will, over parameter, make that the express server run over a username and pa
 
 If no outputFileName is provided it will pipe the standard output stream `process.stdout`
 
-## User the ssr output as stream
+## Use the ssr output as stream
 
-It use the stdout stream so you can do things like:
+It uses the stdout stream so you can do things like:
 
 ```ssh
   $ sui-ssr archive > ./myFileNameOrWhatever.zip
@@ -72,9 +72,9 @@ It use the stdout stream so you can do things like:
 
 ## Hooks
 
-If you want change the server´s behavior for very specific business operation like handler error or logging. You must to use the hooks concept.
+If you want to change the server´s behavior for very specific business operation like handling errors or logging you must use hooks.
 
-To do that define a file `src/hooks.js` this file must be something like this:
+To do that define a file `src/hooks.js` which will look as follows:
 
 ```
 import TYPES from '@s-ui/ssr/hooks-types'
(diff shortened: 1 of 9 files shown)


packages/sui-ssr/README.md
```diff
@@ -87,12 +87,24 @@ export default {
...
+Here we implement a direct log into the console. Each hook type could be a middleware function or an array of middlewares functions.
 
-You can checks with hooks there are available in the hooks-types.js file.
+You can check which hooks are available in the hooks-types.js file.
 
 There are two default hooks for 404 and 500 errors. both will look for a 404.html or 500.html file in the src folder and show this file. If you dont define this files, you will get a generic error page.
 
+## Environment variables
+
+You can define environment variables by creating a yml file called `public-env.yml` in your SPA root directory:
```
carlosvillu: Add a warning to avoid commit this file to the repo.

packages/sui-ssr/README.md
```diff
@@ -87,12 +87,25 @@ export default {
...
```
davidbarna: naming could `env.yml`. It's already in the public repo. Or `spa-env.yml` if it's only for SPAs

davidbarna:
Your build is failing because of new sass-lint rules.

davidbarna:
Errors were fixed here: https://github.com/SUI-Components/sui/pull/379
You can just rebase from master

miduga:
What's the benefit of using yaml instead a json object? I see more natural use a JSON object for this and avoid loading a library for parsing it. Is any special reason?

betobe:
> What's the benefit of using yaml instead a json object? I see more natural use a JSON object for this and avoid loading a library for parsing it. Is there any special reason?

YAML was just of my preference because it is robust, supports more features than json (just a nice-to-have) and, in my opinion, I think it is more readable. I talked to some colleagues in my team and they think the same. I see no cost in using a third party library for parsing a yaml file as it doesn't really have any performance impact because it is loaded once when the app starts and it's on the server side so no bundling cost neither. About the "natural" aspect, I guess you refer to the common file formats a JS app work with, but from a devops perspective we think slightly the opposite as a YAML is a nicer format to deal with.

I think it is a matter of personal opinions. Let me know if you are ok with it so I can close the PR. Any thoughts?

miduga:
I'm ok to merge it knowing that's more a personal opinion than having some specific reason behind it.

I don't see any special feature that you need in this case and it seems that having to install a dependency for something that you just could do native... It's fine embracing the devops perspective but I would be more concerned about the devs that will use it. For example, the ecosystem was intended to support JS files as we don't have linter of YAML files and one space will break the file, while we can have JS files easily check and it will throw an error to avoid this kind of problem.

Anyway, I don't have a strong opinion about it, if you are happy with YAML, it's fine for me. :+1:

PR #377, opened 2018-09-05
fix(sui-i18n): fixed a bug related with regex

carlesnunez:
## Description
I've changed the REGEX in order to fix an edge case that was throwing errors. 

I've also increased our test suite in order to avoid future problems regarding slugify function.

### DIFF
--- a/packages/sui-i18n/src/slugify.js
+++ b/packages/sui-i18n/src/slugify.js
@@ -11,7 +11,7 @@ export function slugify(str) {
   return str
     .toLowerCase()
     .replace(/.{1}/g, replaceCharIfNeeded)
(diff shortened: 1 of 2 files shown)


packages/sui-i18n/src/slugify.js
```diff
@@ -11,7 +11,7 @@ export function slugify(str) {
   return str
     .toLowerCase()
     .replace(/.{1}/g, replaceCharIfNeeded)
-    .replace(/[^a-z0-9 - .]/g, '') // remove invalid chars only letters numbers, dashes and dots
+    .replace(/[^a-z0-9. -]/g, '') // remove invalid chars only letters numbers, dashes and dots
```
carlosvillu: You have to escape the dot

Because dot means all characters

PR #381, opened 2018-09-07
feat(sui-lint): throw error if only in tests and use a const for rule values

miduga:
- [x] Add rule to avoid sending `.only` tests to be commited to the repo:
image
- [x] Use a dictionary in order to add the rules values.


packages/sui-lint/package.json
```diff
@@ -25,6 +25,7 @@
     "eslint-config-standard-react": "6.0.0",
     "eslint-plugin-chai-friendly": "0.4.1",
     "eslint-plugin-import": "2.10.0",
+    "eslint-plugin-no-only-tests": "^2.0.1",
```
jcger: you forgot the caret

PR #382, opened 2018-09-13
[sui-ssr] Add Device to context

carlosvillu:


packages/sui-ssr/build-device.js
```diff
@@ -0,0 +1,8 @@
+export const buildDeviceFrom = ({request = {}, window = {}}) => {
+  return {
+    // https://tc39.github.io/proposal-optional-chaining/
+    'user-agent':
```
miduga: Why not returning it camelCase?

carlosvillu: No reason, just because is the header´s name

carlesnunez: If is not a huge change or painful I'll change it by camelCase in order to be more compliant with the var names standards

packages/sui-ssr/build-device.js
```diff
@@ -0,0 +1,8 @@
+export const buildDeviceFrom = ({request = {}, window = {}}) => {
```
jcger: IMO something like `buildDeviceInfo`, `getDeviceData` or even just `buildDevice` would be more descriptive

carlosvillu: The idea is combine the name of the method with the param name, to get something like: `buildDeviceFrom({window})`

To achive a more semantic method call

PR #385, opened 2018-09-17
fix(sui-deploy): added branch name in each travis build to deploy our current branch in NOW with his own name.

ddmarin94:
Travis has changed the way into checkout to other branches in builds. 
- Before this change we were in the current branch of the PR. Then we knew the branch name and we could use it.
- After this change, is a little bit more complicated take the current branch name to use it as url. First of all, we need to use a new environment variable name process.env.TRAVIS_PULL_REQUEST_BRANCH.

As a fallback, we will have process.env.TRAVIS_BRANCH and getGitBranch func.

### DIFF
--- a/packages/sui-deploy/src/utils.js
+++ b/packages/sui-deploy/src/utils.js
@@ -19,7 +19,10 @@ const getDeployNameFromProgram = async program => {
   const deployName =
     deployBaseName +
     (program.branch
-      ? '-' + (process.env.TRAVIS_BRANCH || (await getGitBranch()))
+      ? '-' +
+        (process.env.TRAVIS_PULL_REQUEST_BRANCH ||
+          process.env.TRAVIS_BRANCH ||
+          (await getGitBranch()))
       : '')
   return toKebabCase(deployName)
 }


packages/sui-deploy/src/utils.js
```diff
@@ -19,7 +19,10 @@ const getDeployNameFromProgram = async program => {
   const deployName =
     deployBaseName +
     (program.branch
-      ? '-' + (process.env.TRAVIS_BRANCH || (await getGitBranch()))
+      ? '-' +
```
carlesnunez: Why instead of couple our tool with 3rd party integrations we don't decide to pass the 'deploy alias' by param?

Something like `sui-deploy ... ... --deployName=$TRAVIS_PULL_REQUEST`

Doing that we are avoiding be coupled with TRAVIS on an open source tool.

What do you think?

carlesnunez: @ddmarin94 Talking with carlos we decided to add a CLI option in order to allow our users to decide which name of branch are they going to use as I added. In order to maintain the retrocompatibility with our current users we will don't remove the travis env var.

davidbarna: The first arg is already the deployName...
Guys, this "option" already exists:
```sh
sui-deploy "milanuncios-$TRAVIS_PULL_REQUEST" --now
```

davidbarna: Plus, if you have to to sonething, move `process.env.TRAVIS_BRANCH` logic into getGitBranch func...

carlesnunez: Wow you are right! So sorry about that. I am agree to remove the travis branch in order to pass it as an option

ddmarin94: This pr will be closed, we found a different way to solve this problem.

PR #386, opened 2018-09-20
[sui-bundler] Fix webpack mode when compile a server

carlosvillu:


packages/sui-bundler/webpack.config.server.js
```diff
@@ -5,6 +5,7 @@ const babelRules = require('./shared/module-rules-babel')
 
 let webpackConfig = {
   context: path.resolve(process.cwd(), 'src'),
+  mode: process.env.NODE_ENV === 'production' ? 'production' : 'development',
```
Hyunk3l: Don't know how to achieve it, but I'd allow more environments (testing, staging, preprod..whatever)

carlosvillu: Webpack only allow three value in this key: 'production', 'development' and 'none'

But this key is about how Webpack compile the assets, not the environment where the server is running

PR #392, opened 2018-10-01
Feature/pass env vars to now deployer

carlesnunez:
## Description
This PR allows our tool to receive -e option in order to do deploys with environment vars to NOW.sh or the deployer that we setted up.

## Example
More info on readme

### DIFF
--- a/packages/sui-deploy/src/client/now.js
+++ b/packages/sui-deploy/src/client/now.js
@@ -7,13 +7,18 @@ const DEFAULT_DIR = join(process.cwd(), 'public')
 const NOW_VERSION = '11.2.8'
 
 // Get args of `now` command according to params
-const getNowCommandArgs = ({name, dir, auth, token}) => {
+const getNowCommandArgs = ({name, dir, auth, token, environmentVars}) => {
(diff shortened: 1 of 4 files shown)


packages/sui-deploy/src/client/now.js
```diff
@@ -7,14 +7,19 @@ const DEFAULT_DIR = join(process.cwd(), 'public')
...
+const getNowCommandArgs = ({name, dir, auth, token, environmentVars}) => {
   const args = [dir, '--name=' + name, `-t "${token}"`]
   if (auth) {
     const [user, password] = auth.split(':')
     user &&
       password &&
       args.push(`-e SERVE_USER='${user}'`, `-e SERVE_PASSWORD='${password}'`)
   }
+  if (environmentVars) {
+    environmentVars.forEach(ENVIRONMENT_VAR =>
```
jcger: we usually use uppercase for variables that are hardcoded

carlesnunez: But this ones are not hardcoded per se am I right? This ones are introduced by the user from prompt.

jcger: 🤔 but it can be anything

PR #397, opened 2018-10-03
feat(sui-studio): remove all about compliant in sui-studio as not longer necessary

miduga:
Remove all about compliant components as is not longer necessary.


carlosvillu:
@jcger Could we close this PR ?!

PR #398, opened 2018-10-05
feat(sui-studio): generate kebab case package name

jcger:
npm package names must be kebab case

```
> sui-studio generate testCategory testName 
```
was setting `prefix/scope-testCategory-testName` as package name


carlesnunez:
This was really painful this fix will be awesome!!

PR #410, opened 2018-10-17
Feat/sui js

ddmarin94:
## Description
Created utils in order to have camelcase and capitalcase strings transformations.


packages/sui-js/src/string/camel-case.js
```diff
@@ -0,0 +1,5 @@
+export const toCamelCase = string =>
```
jcger: Please add tests

PR #412, opened 2018-10-23
feat(sui-polyfills): add new polyfill for IE11 in order to let .remove method to work on Elements

miduga:
In order to add the possibility to use `Element.remove` in our developments as it's not supported by IE11. Alternatively, we need to use `element.parentNode.removeChild(element)` everytime we need to remove an element from the DOM.


tureey:
@miduga could you please share how did you solve it? We still have this problem. Would be nice if we could fix it quickly!

PR #413, opened 2018-10-23
Fix/ssr build device

ddmarin94:
## Description
Quick fix, passed from es6 to es5 haven't being transpiled in our web app and crashes in IE11.


packages/sui-ssr/build-device.js
```diff
@@ -1,8 +1,13 @@
-export const buildDeviceFrom = ({request = {}, window = {}}) => {
+export var buildDeviceFrom = function(params) {
+  var win = params.window
+  var request = params.request
+  request = request === undefined ? {} : request
```
carlesnunez: Am I right that we can change it by this?

var request = params.request || {}

Same with window? Or the undefined explicit comparision have a purpose?

miduga: The explicit comparision is to keep the exact same behavior as before for default value in param function. The one you're mentioned is a loose, as is checking a falsy value instead. In my opinion, both could work for this case.

PR #414, opened 2018-10-24
fix(sui-widget-embedder): update bundler version

kikoruiz:
## Description
We need to update the bundler version in order to avoid multiple webpack runtimes collision, if applies.

## Related Issue
https://github.com/SUI-Components/sui/pull/400


miduga:
And is that all?

kikoruiz:
PR merged by mistake. Code from master has been removed.

PR #420, opened 2018-11-05
fix(sui-widget-embedder): update files to load

jcger:


packages/sui-widget-embedder/downloader/index.js
```diff
@@ -85,22 +84,21 @@
...
-    var baseUrl = cdn + '/' + page + '/'
     var manifest = manifests[page]
     var loadStyle = function() {
       var css = manifest['app.css'] || manifest['main.css']
-      return css ? load.css(baseUrl + css) : Promise.resolve()
+      return css ? load.css(css) : Promise.resolve()
     }
     var loadScripts = [
-      manifest['runtime.js'],
+      manifest['runtime~app.js'],
```
jcger: I'm not sure if this will always be the case, do someone know?

jcger: I updated this PR in order make the entry point always to be called app so this should always be the same

PR #422, opened 2018-11-06
Feat/sui svg

jcger:


packages/sui-svg/package.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "dependencies": {
+    "@s-ui/component-peer-dependencies": "latest",
+    "@compositor/kit": "1.0.47",
+    "@s-ui/bundler": "3",
+    "@s-ui/helpers": "1",
+    "@s-ui/js": "1",
+    "@svgr/core": "3.1.0",
+    "commander": "2.19.0",
+    "styled-components": "4.0.2",
```
miduga: ???

carlosvillu: It is a dependency of Kit

miduga: A peer dependency? :/ We might want to re-check this in the future and extract the behaviour that compositor is giving to us as this doesn't seem to be reasonable.

PR #430, opened 2018-12-05
fix(sui-widget-embedder): disable zindex optimization on cssnano

carlesnunez:
## Description
Once I've been working with widgets integrating the IAB modal I've seen that even if I was forcing z-index to high values my final result bundled was always less than the number that I was setting (for example if I setted z-index: 2500 my bundled value was 2)

This is caused by a default behavior of cssnano (more info here):
https://github.com/ben-eb/gulp-cssnano/issues/8

In fact we are not using css nano directly but we are using optimize-css-assets-webpack-plugin which is using cssnano as a dependency.

In order to fix that ONLY ON THE WIDGETS I've decided to modify the widget package as if I modify the webpack.conf of the bundler we could step into unexpected behaviors on all the sites that are using s-ui/bundler.

### DIFF
--- a/packages/sui-widget-embedder/compiler/production.js
+++ b/packages/sui-widget-embedder/compiler/production.js
@@ -1,6 +1,8 @@
 const path = require('path')
 const webpack = require('webpack')
 const prodConfig = require('@s-ui/bundler/webpack.config.prod')
+const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin')
+const uglifyJsPlugin = require('@s-ui/bundler/shared/uglify')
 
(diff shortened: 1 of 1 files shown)


packages/sui-widget-embedder/compiler/production.js
```diff
@@ -1,6 +1,8 @@
 const path = require('path')
 const webpack = require('webpack')
 const prodConfig = require('@s-ui/bundler/webpack.config.prod')
+const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin')
```
JoanClaret: hey! s
hould we install it in the `package.json` file or it's already available?

PR #441, opened 2018-12-13
Feature/move to npm 6 and latest node version

midudev:
Use latest node and npm versions for travis and inform about this in the CONTRIBUTING files.

### DIFF
--- a/packages/sui-dashboard/src/index.js
+++ b/packages/sui-dashboard/src/index.js
@@ -1,5 +1,4 @@
 const fg = require('fast-glob')
-const {promisify} = require('util')
 
 const flat = arr => [].concat(...arr)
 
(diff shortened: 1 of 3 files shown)


packages/sui-dashboard/src/index.js
```diff
@@ -66,31 +65,3 @@ module.exports.stats = async ({repositories, root, dry}) => {
     }
   }
 }
-
-// Dead code, but I dont want to remove
```
carlosvillu: 😭😭😭😭😭😭

midudev: xDDDDDDD

midudev: It will remain forever, in this PR.

PR #444, opened 2018-12-26
Sui hoc no default value

juanmaguitar:
No default value or propType definition for value (`any`) so `withStateValue` can be used on any type of data (string, array,...)

### DIFF
--- a/packages/sui-hoc/package.json
+++ b/packages/sui-hoc/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@s-ui/hoc",
-  "version": "1.5.0",
+  "version": "1.6.0",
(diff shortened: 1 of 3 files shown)


packages/sui-hoc/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@s-ui/hoc",
-  "version": "1.5.0",
+  "version": "1.6.0",
```
jcger: did you already release it or why is this + changelog here?

juanmaguitar: I'm not sure about this. The only explanation I cant think of regarding this is that i took these changes when merging from master to get the latest changes on this PR

PR #445, opened 2018-12-27
feat(sui-js): added pipe method

AgonisticKatai:
Added 'pipe' method to sui-js module


packages/sui-js/README.md
```diff
@@ -5,6 +5,22 @@
...
 ```
 
+## pipe
+Consist of a chain of processing functions, where the output of each element is the input of next
+
+```js
+import pipe from @s-ui/js/lib/pipe
+
+execute = async search => {
+  const pageName = await this.pipe(
```
jcger: it should be without `this`, right?

packages/sui-js/src/pipe/index.js
```diff
@@ -0,0 +1,4 @@
+const pipe = (...funcs) => arg =>
+  funcs.reduce((value, func) => func(value), arg)
```
jcger: it would be nice if you could add some tests

jcger: I would rename `(value, func)` into `(acc, value)` for easier reading as so are called the args passed through the reducer fn, up to you

packages/sui-js/README.md
rmoralp: Could you remove `this.`from here?. Just a detail to call pipe import

packages/sui-js/README.md
```diff
@@ -5,6 +5,22 @@
...
+    this._fromQueryStringToQueryParamsPageMapper.map,
+    this._fromQueryParamsToTargetPageValueObjectPageMapper.map
+  )(search)
+
+  return pageName
+}
```
jcger: not sure if this is a good example as only those familiar with our domain stuff will know what that mapper functions should look like and what search is. It would be better if you could add an example that works by copy and paste it

packages/sui-js/README.md
AgonisticKatai: ups, sorry, when I created the method, I had the function in the same file and it stayed there by mistake

AgonisticKatai: ups, sorry, when I created the method, I had the function in the same file and it stayed there by mistake

packages/sui-js/src/pipe/index.js
AgonisticKatai: ok thx!!

packages/sui-js/README.md
AgonisticKatai: It's true, I change it! Thank you!

Issue #446, opened 2019-01-03
[sui-test] Add an option to set a complete UA


kikoruiz:
### Package

[sui-test]@[1.14.0]

### Description

We're missing the possibility to set to Cypress config a complete User-Agent, not only append a substring into a default one.

PR #448, opened 2019-01-03
feat(sui-test): add overwrite user agent option

rmoralp:
--userAgent option has been added
It is required in order to change the entire user agent.  Alto, --userAgentAppend behavior stays the same.

Related Issue: Fix [#446](https://github.com/SUI-Components/sui/issues/446)

**To test:**
1.  ```cd frontend-mt--e2e-tests```
2. ```node ../sui/packages/sui-test/bin/sui-test.js e2e --baseUrl="http://pre.coches.net"  --scope="coches" --gui --userAgent="Screaming Frog SEO Spider/8.3"```
3. Then, run a test and write ```navigator.userAgent``` in console

### DIFF
--- a/packages/sui-test/bin/sui-test-e2e.js
+++ b/packages/sui-test/bin/sui-test-e2e.js
@@ -37,19 +37,32 @@ program
     '-U, --userAgentAppend <userAgentAppend>',
     'Append string to UserAgent header.'
   )
+  .option(
+    '-UA, --userAgent <userAgent>',
+    'Overwrite string to UserAgent header.'
+  )
   .option('-s, --scope <spec>', 'Run tests specifying a subfolder of specs')
   .option('-G, --gui', 'Run the tests in GUI mode.')
   .on('--help', () => console.log(HELP_MESSAGE))
   .parse(process.argv)
 
-const {baseUrl, userAgentAppend, gui, screenshotsOnError, scope} = program
+const {
+  baseUrl,
+  userAgentAppend,
+  userAgent,
+  gui,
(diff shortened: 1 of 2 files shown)


packages/sui-test/bin/sui-test-e2e.js
```diff
@@ -37,19 +37,32 @@ program
...
+} = program
 const cypressConfig = {
   integrationFolder: path.join(TESTS_FOLDER, scope || ''),
   baseUrl,
   fixturesFolder: path.join(TESTS_FOLDER, 'fixtures')
 }
 
-if (userAgentAppend) {
+if (userAgent) {
+  cypressConfig.userAgent = `"${userAgent}"`
```
kikoruiz: This can be simplified with: `cypressConfig.userAgent = userAgent`

rmoralp: You're right but, At the beginning I did it but I see that the function `objectToCommaString` sets the userAgent as `userAgent: 'undefined'` if you don't set --userAgend or --userAgentAppend because it will assign undefined in this property. So, I prefer not to modify `objectToCommaString`

rmoralp: If cypress.json has userAgent: 'undefined', Cypress browser will has a 405 - Method Not Allowed (text/html) response

kikoruiz: Yes, sorry, I was wrong, the quotes are needed in this case ;) I had the same issue early this morning 😛

PR #458, opened 2019-01-11
feat(sui-js): added helpers needed for several sui-components

juanmaguitar:


packages/sui-js/src/dom/index.js
```diff
@@ -0,0 +1,10 @@
+export const getCurrentElementFocused = () => document.activeElement
+
+export const getFocusedItemIndex = domItems => {
+  const currentElementFocused = getCurrentElementFocused()
+  return domItems.reduce(
+    (focusedItemIndex, item, index) =>
```
jcger: This function seems strange to me as you are returning an index using the reduce method but without accumulating anything, isn't this the same?

```js
Array.from(options).findIndex(item => item === currentElementFocused)
```

jcger:
please add some tests

Issue #432, opened 2018-12-10
SSR to return the SPA in case of error


betobe:
### Package

sui-ssr@2.11

### Description

To return the app file (index.html) with the corresponding error status as header and also as part of the app data, instead of reading and returning a 4xx|5xx.html file when an error occurs.
+ This would make the SPA capable of choosing what to do in case of any error (show a message; redirect to somewhere else; offer some other content).
+ Users would be able to continue using the app even in case of error.
+ Errors pages would be built as react components (as a regular page).
+ The need of "one file per error" and its maintenance would be removed.
+ Given that New Construction SPA works for different sites with a custom look and feel for each site ( and site-config identification is done by the SSR) this feature would make error pages to be compliant with each site brand look and feel straightforward.

⚠️  This change would not be backward compatible.

carlosvillu:
Why dont you just copy your public/index.html to public/400|500.html ?!

carlosvillu:
Or maybe you can use the hooks, to response what you want.

betobe:
> Why dont you just copy your public/index.html to public/400|500.html ?!
> Or maybe you can use the hooks, to response what you want.

Although It might be possible to find a way of emulate what I need with a hook, it would be a workaround. I want to avoid workarounds with this proposal. What are your concerns about this approach?

carlosvillu:
The main idea of sui-ssr is been 100% agnostic to the SPA. But NC SP/SSR are not agnostic anymore. I prefer to use a workaround for a specific project is that is able than change the generic behavior in the rest of the SPAs. 

about onstart-hook, I prefer to avoid fully extendable code, to be able to evolve the tools with the minimal risks.

betobe:
Well, everything I've suggested (this proposal + open hooks) is in favor of agnosticity (plus benefits already listed in the proposal).

Also,
> The main idea of sui-ssr is been 100% agnostic to the SPA. But NC SP/SSR are not agnostic anymore.

I think It never has and it never will. SSR knows a lot about SPA and it is just fine. But I really don't understand why in your opinion NC case is not agnostic based on your definition of agnosticity and the rest of SPAs are indeed. Could you clarify this?

Let me explain myself. This is what I see (and it is a quick shallow list):

PROPOSAL + new hooks system:
**spa-knowing-ssr**:
- hooks implementation
- ~~xxx.html files~~ current

**ssr-knowing-spa**:
- spa's hooks implementation magic import
- spa's node_modules magic imports
- spa's index.html file
[...]

PR #462, opened 2019-01-14
feat(sui-ssr): delay early flush in order to properly handle getInitialProps exception

betobe:
* To be able of setting the right http status code (404; 500; 301) by using a hook defined in the SPA.

## Description
Early flush was done before knowing the right HTTP code for the response. This generates soft 404s in case of any error.

Now the server will wait for getInitialProps result before flushing. This may increase the TTFB but technically is the right way to go as the SSR should not be responsible for the performance of the service consumed by getInitialProps.

Also, with this change, getInitialProps may return an object with a `.error` property in order to be managed by a custom hook if needed. E.g.: 
- getInitialProps returns Error(404)
- A custom hook is implemented for reading that message and decide what to do at server level (e.g.:load index.html and return 404 http code) and app level (e.g.: Display 404 message and/or redirect to some other page based on the error)

Related issue https://github.com/SUI-Components/sui/issues/432

### DIFF
--- a/packages/sui-ssr/server/ssr/index.js
+++ b/packages/sui-ssr/server/ssr/index.js
@@ -70,22 +70,35 @@ export default (req, res, next) => {
       const device = buildDeviceFrom({request: req})
       const [criticalHTML, bodyHTML] = html.split('</head>')
 
-      res.type('html')
-      res.write(criticalHTML)
-      res.flush()
-
       const context = await contextFactory(
         createServerContextFactoryParams(req)
       )
-      const {
-        initialProps,
-        reactString,
-        performance
-      } = await ssrComponentWithInitialProps({
-        context: {...context, device},
-        renderProps,
-        Target: withContext(context)(RouterContext)
-      })
+
+      let initialData
+
+      try {
+        initialData = await ssrComponentWithInitialProps({
+          context: {...context, device},
+          renderProps,
+          Target: withContext(context)(RouterContext)
+        })
+      } catch (err) {
+        // Managing 500 exception
+        return next(err)
+      }
+
+      const {initialProps, reactString, performance} = initialData
+
+      if (initialProps.error) {
+        // getInitialProps may return an error as property
+        // This error can be handled by the default hook or by a custom hook
+        return next(initialProps.error)
+      }
+
+      // Early flush
+      res.type('html')
+      res.write(criticalHTML)
+      res.flush()
 
       // res.set({
       //   'Server-Timing': `


carlosvillu:
Do you have an Idea of how much this PR will impact the performance of our TTFB ?!

packages/sui-ssr/server/ssr/index.js
```diff
@@ -70,22 +70,35 @@ export default (req, res, next) => {
...
+          Target: withContext(context)(RouterContext)
+        })
+      } catch (err) {
+        // Managing 500 exception
+        return next(err)
+      }
+
+      const {initialProps, reactString, performance} = initialData
+
+      if (initialProps.error) {
```
carlosvillu: Maybe we could use a less common name to this. What do you think about use: `__error__` like key ?!

carlosvillu: Or even be more explicit and have to import something from sui-ssr like we did with the hooks.

betobe: It's a common name because it is an Error instance indeed. I can't imagine a simpler and clearer contract. But I'm open to change it if you have a clear suggestion about the name. Regarding hooks, If this contract is clear and it's not prone to change, I don't really see the point of implementing it in some other way. Thoughts?

midudev:
I think this is the correct way to do and that @betobe is right when states that the SSR shouldn't be concerned about backend services. That was a patch in order to avoid very slow backend services to break the TTFB but still is a patch that's holding us back to properly managing errores from the server.

We might want to push forward more CSR in order to not be concerned about this anymore.

But we might want to consider if it could be a good idea to, at least, keep the old behaviour as default and add this one behind a flag.

carlosvillu:
Have a lot of sense what you say, And for example tomorrow we have a meeting in MA to use Dynamic Rending.

I think that your idea of using a flag it is fine. Please leave the current behavior as default and add a flag to apply the early flush.

packages/sui-ssr/server/ssr/index.js
midudev: `getInitialProps` already returns a Promise. Why do not use the `catch` in order to handle the error instead having to return an error property? Maybe some changes are needed, or not, in react-initial-props package but it sounds like it makes more sense.

midudev: In fact, the error of the `getInitialProps` could be captured here: https://github.com/SUI-Components/sui/pull/462/files/3eb50222dea1bbc6eafb7ce165f336acd9170605#diff-c1aac530bdd240d83babeab7fefaf59aR86 and no `error` property would be needed.

carlosvillu: I am thinking that maybe this place is not the best place to do that. Because what we want is to avoid is render our page if we have an Error in the getInitialprops function. But the current GIP package will pass the error prop to our page component. Is that what we want ?!

betobe:
@carlosvillu, @midudev Besides the early-flush flag I've also add a flag for allowing the SPA to display 404 errors.

About minors and majors:
Technically I should (and I would prefer to) publish a major version of the SSR because of the changes in a default hook. What do you think? On the other hand, I think I could publish a minor version for sui-react-initial-props but I'm not 100% sure. Thoughts?

PR #464, opened 2019-01-15
feat(sui-bundler): Add sourcemaps support to improve sentry integration

igncalleja:
## Description
This PR adds sourcemap generation support for sui-bundler. It expects a new section `sourcemaps` in the sui-bundler project package.json configuration to get from there the desired devtool webpack value. By default, the behaviour would be same as it is now and no sourcemaps are generated if this section is not present.

A new source-map-loader has been added for all js/jsx files in the babel rules config and the flag for sourcemaps in the uglifyJS plugin config has been changed.

## Example
```json
{
  "sui-bundler": {
    "sourcemaps": {
      "dev": "cheap-module-eval-source-map",
      "prod": "hidden-source-map"
    }
  }
}
```

### DIFF
--- a/packages/sui-bundler/shared/uglify.js
+++ b/packages/sui-bundler/shared/uglify.js
@@ -3,5 +3,5 @@ const UglifyJsPlugin = require('uglifyjs-webpack-plugin')
 module.exports = new UglifyJsPlugin({
   cache: true,
   parallel: true,
-  sourceMap: true // set to true if you want JS source maps
+  sourceMap: true
 })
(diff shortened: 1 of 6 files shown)


packages/sui-bundler/shared/uglify.js
```diff
@@ -3,5 +3,5 @@ const UglifyJsPlugin = require('uglifyjs-webpack-plugin')
 module.exports = new UglifyJsPlugin({
   cache: true,
   parallel: true,
-  sourceMap: false // set to true if you want JS source maps
+  sourceMap: true // set to true if you want JS source maps
```
jcger: you could remove the comment

PR #469, opened 2019-01-18
[sui-ssr] send Content-Type header setup to HTML

carlosvillu:


packages/sui-ssr/server/dynamicRendering/index.js
```diff
@@ -14,6 +14,8 @@ export default function dynamicRendering(fallback, dynamicsURLS = []) {
       req.url.match(new RegExp(url))
     )
 
+    enabledDynamicRendering && resp.type('html')
```
betobe: You could replace the ternary op and use an if..else in line 19 so that you can add `resp.type('html')` there.

PR #470, opened 2019-01-18
feat(sui-hoc): intersection-observer hoc

juanmaguitar:


packages/sui-hoc/package.json
```diff
@@ -14,6 +14,9 @@
     "prop-types": "15",
     "react": "15 || 16"
   },
+  "dependencies": {
+    "intersection-observer": "0"
```
jcger: please use the exact version

packages/sui-hoc/src/withIntersectionObserver.js
```diff
@@ -0,0 +1,48 @@
+/* global IntersectionObserver */
+import React, {Component} from 'react'
+
+if (typeof window !== 'undefined' && window.document) {
```
midudev: What about using a dynamic import only if the browser doesn't have support for it. So the check would be done in the componentDidMount, which is only executed in the client.

juanmaguitar: With this code 
```
if (typeof window !== 'undefined' && window.document) {
  require('intersection-observer')
}
```
we assure we only load the library/pollyfill in the client. 

For what I've seen in other implementations in Schibsted, the general criteria is to load always the pollyfill and let that code to check if the specific feature is available and apply proper changes if not

What exactly is your suggestion here?

juanmaguitar:
@midudev I have added the code to export also the version that will allow the developer to customize the intersection observer w/ custom options

packages/sui-hoc/src/withIntersectionObserver.js
carlosvillu: using that IF you will end with the polyfill in your bundler. You have to use the `import()` statement to have lazy loading of this polyfill.

PR #471, opened 2019-01-22
feat(sui-widget-embedder): allow to add a blacklist of reg exps

kikoruiz:
## Description
Regarding the best practice of loading one widget bundle by page, in the most cases we'll need to load that bundle on a specific number of pages, via the existing param `pathnameRegExp`, but sometimes we'll want to load it everywhere but not on those pages, so that, the param `blacklistedRegExps` will help us to skip them.

## Example
```js
// pages/all-not-static.js
{
  "version": "1.0.0",
  "private": true,
  "pathnameRegExp": ".*",
  "blacklistedRegExps": [
    "contact.html",
    "about.html"
  ]
}
```

### DIFF
--- a/packages/sui-widget-embedder/downloader/index.js
+++ b/packages/sui-widget-embedder/downloader/index.js
@@ -115,9 +115,9 @@
     if (
       blacklistedRegExps &&
       blacklistedRegExps.length > 0 &&
-      !blacklistedRegExps.some(regExp =>
-        window.location.pathname.match(new RegExp(regExp))
-      )
+      !blacklistedRegExps.some(function(regExp) {
+        return window.location.pathname.match(new RegExp(regExp))
+      })
     ) {
       pages.push(page)
     } else if (window.location.pathname.match(new RegExp(pathnameRegExp))) {
--- a/packages/sui-widget-embedder/downloader/index.js
+++ b/packages/sui-widget-embedder/downloader/index.js
@@ -114,6 +114,7 @@
 
     if (
       blacklistedRegExps &&
+      blacklistedRegExps.length > 0 &&
       !blacklistedRegExps.some(regExp =>
         window.location.pathname.match(new RegExp(regExp))
       )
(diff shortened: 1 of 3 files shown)


packages/sui-widget-embedder/downloader/index.js
```diff
@@ -108,11 +108,22 @@
...
-  for (var key in pathnamesRegExp) {
-    if (window.location.pathname.match(new RegExp(pathnamesRegExp[key]))) {
-      pages.push(key)
+  for (var page in pageConfigs) {
+    var blacklistedRegExps = pageConfigs[page].blacklistedRegExps
+    var pathnameRegExp = pageConfigs[page].pathnameRegExp
+
+    if (
+      blacklistedRegExps &&
+      !blacklistedRegExps.some(regExp =>
```
midudev: Arrow function might not be compatible with all browsers. Let's use a normal one.

midudev: Extracting the function for checking regExp.
```
function matchPathnameWithRegex(regExp) {
  return window.location.pathname.match(new RegExp(regExp)
}
```

Later:
```
!blacklistedRegExps.some(matchPathnameWithRegex)

...

else if (matchPathnameWithRegex(pathnameRegExp)
```

packages/sui-widget-embedder/downloader/index.js
```diff
@@ -108,11 +108,23 @@
   }
 
   var pages = []
-  for (var key in pathnamesRegExp) {
-    if (window.location.pathname.match(new RegExp(pathnamesRegExp[key]))) {
-      pages.push(key)
+  for (var page in pageConfigs) {
+    var blacklistedRegExps = pageConfigs[page].blacklistedRegExps
```
midudev: In order to avoid checks with the if, we might want to give a default value.
```
var blacklistedRegExps = pageConfigs[page].blacklistedRegExps || []
```

That would simplify the if a little bit.

packages/sui-widget-embedder/downloader/index.js
kikoruiz: Thanks for your suggestions @midudev!!

Issue #488, opened 2019-02-13
[sui-js] pipe async functions


rmoralp:
### Package
sui-js@1.20.0

### Description

pipe function does not accept async functions.

PR #490, opened 2019-02-14
Fix/pipe async functions

rmoralp:
Motor domain requires a pipe with promise functions support in order to pipe async mappers

## Related Issue
#488

### DIFF
--- a/packages/sui-js/src/pipe/index.js
+++ b/packages/sui-js/src/pipe/index.js
@@ -1,3 +1,5 @@
 const pipe = (...funcs) => arg => funcs.reduce((acc, value) => value(acc), arg)
-
 export default pipe
+
+export const asyncPipe = (...funcs) => arg =>
+  funcs.reduce((acc, value) => acc.then(value), Promise.resolve(arg))
--- a/packages/sui-js/test/asyncPipeSpec.js
+++ b/packages/sui-js/test/asyncPipeSpec.js
@@ -12,10 +12,12 @@ describe('@s-ui/js', () => {
       const result = await asyncPipe(textToUpperCase, textToArray)(title)
       expect(result).to.be.an('array')
     })
+
     it('should be the same length as the title ', async () => {
       const result = await asyncPipe(textToUpperCase, textToArray)(title)
       expect(result).to.have.lengthOf(title.length)
     })
+
     it('should return the title in camelcase and array form ', async () => {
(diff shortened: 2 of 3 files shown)


jcger:
Please add documentation

packages/sui-js/src/pipe/index.js
```diff
@@ -1,3 +1,5 @@
 const pipe = (...funcs) => arg => funcs.reduce((acc, value) => value(acc), arg)
-
 export default pipe
+
+export const asyncPipe = (...funcs) => arg =>
```
jcger: I don't really understand the point of using this "helpers", don't you know your functions?

I mean, in case of `asyncPipe[bla, bla2, bla3]` and `bla().then(bla2).then(bla3)` everybody will have to check what asyncPipe is the first time but everybody knows what the second example does. Also, in case something changes and you need to add a second value to any of these functions you'r screwed and have to use thens anyways, don't you? Maybe I'm missing something

packages/sui-js/test/asyncPipeSpec.js
```diff
@@ -0,0 +1,27 @@
+/* eslint-env mocha */
+import {expect} from 'chai'
+import {asyncPipe} from '../src/pipe/index'
+
+describe('@s-ui/js', () => {
+  const textToUpperCase = text => text.toUpperCase()
```
jcger: shouldn't you use async functions? what's the point of pipe if this function works with sync function too?

rmoralp:
Readme docs added

packages/sui-js/test/asyncPipeSpec.js
rmoralp: I changed it. asyncPipe accept async and sync functions

packages/sui-js/src/pipe/index.js
rmoralp: The result of `asyncPipe[bla, bla2, bla3]` and `bla().then(bla2).then(bla3)` is the same, and you are right that people may check what asyncPipe do. Nevertheless, In motor we need to chain some fn (async mappers) and it improves the code readability. @carlosvillu suggest us to make this async pipe.

naxhh: I would argue that pipe is a very standard function. Everyone knows what a pipe is and how it works. AsyncPipe is just a detail in how it runs a normal pipe. I don't think it would be really hard to grasp.

PR #496, opened 2019-02-15
[babel-preset-sui] Add @babel/plugin-syntax-export-namespace-from

rmoralp:


midudev:
This makes me wonder if it's correct using the `proposal` for the `"@babel/plugin-proposal-export-default-from"` instead using the `syntax` like this one.

carlosvillu:
As far as I know, Webpack works fine this syntax.

midudev:
@carlosvillu if webpack is handling fine this syntax, then we might want to remove this proposal and move to the syntax one. https://github.com/SUI-Components/sui/pull/496/files#diff-7c7d551550d7f45136a1b402a9ccb4efL14

PR #497, opened 2019-02-18
feat(sui-js): added utils to suit class names

ferransimon:
## Description

Added a new util function to be able to easily suit a class name following the convention for baseComponent{-children}-element--modifier

## Example

`
const baseComponent = suitClass('baseComponent')
<div className={baseComponent({element: 'element'})}/>
`


midudev:
A README about how to use it and some examples would be great.

ferransimon:
@midudev @tureey README with examples added. Please, check it out, and, if it looks fine for you, I will merge it. Thanks!

PR #498, opened 2019-02-18
Fix/use new sui test

midudev:
In order to get sui-test working, upgrade it to version 2 that's compatible with babel@7.

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,7 @@
     "@s-ui/lint": "2",
     "@s-ui/mono": "1",
     "@s-ui/precommit": "2",
-    "@s-ui/test": "1",
+    "@s-ui/test": "2",
     "chai": "3.5.0",
     "cross-env": "5.1.6",
     "husky": "0.13.4",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -29,7 +29,7 @@
     "@s-ui/lint": "2",
     "@s-ui/mono": "1",
     "@s-ui/precommit": "2",
-    "@s-ui/test": "1",
+    "@s-ui/test": "2",
```
carlosvillu: I made this change 4 days ago: https://github.com/SUI-Components/sui/commit/257d29d192ddf442e7365d35bc74736fd8ce8903

packages/sui-js/test/pipeSpec.js
```diff
@@ -4,7 +4,7 @@ import pipe from '../src/pipe/index'
 
 describe('@s-ui/js', () => {
   const textToUpperCase = text => text.toUpperCase()
-  const textToArray = text => [...text]
+  const textToArray = text => text.split('')
```
carlosvillu: And change this functions, for others: https://github.com/SUI-Components/sui/commit/fd8df3b27f017b5d790063a060a662d3cbf2d355

But I have to say that your fix is more elegant.

PR #499, opened 2019-02-18
chore(META): show failure npm log on travis

midudev:
After failure, show the log of npm in order to get more insights when the build fails.


midudev:
It doesnt work like that T_T It seems that the process.exit breaks the build.

PR #508, opened 2019-02-20
Fix/sui test on sui

midudev:
- Use sinon 2 for karma-webpack compatibility with sui-test@2.
- Transform export default from as it's not supported by karma-webpack.

### DIFF
--- a/packages/sui-test/bin/karma/config.js
+++ b/packages/sui-test/bin/karma/config.js
@@ -39,7 +39,10 @@ const config = {
                     }
                   ]
                 ],
-                plugins: [require.resolve('babel-plugin-dynamic-import-node')]
+                plugins: [
+                  require.resolve('babel-plugin-dynamic-import-node'),
+                  require.resolve('@babel/plugin-proposal-export-default-from')
+                ]
               }
             }
           ]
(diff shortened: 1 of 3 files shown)


packages/sui-test/bin/karma/config.js
```diff
@@ -39,7 +39,10 @@ const config = {
                     }
                   ]
                 ],
-                plugins: [require.resolve('babel-plugin-dynamic-import-node')]
+                plugins: [
+                  require.resolve('babel-plugin-dynamic-import-node'),
+                  require.resolve('@babel/plugin-proposal-export-default-from')
```
carlosvillu: We have this plugin in Babel-preset-SUI

midudev: The one in the babel-preset-sui is the syntax. (we had the proposal and decided to go only with syntax as is supported by webpack). The problem is karma-webpack doesn't support the same as webpack (they have it fixed but not released yet). For now, jsut, transpile stuff for testing and we will remove this when karma fix it.

packages/sui-test/package.json
```diff
@@ -16,6 +16,7 @@
   "dependencies": {
     "@babel/cli": "7.2.3",
     "@babel/core": "7.2.2",
+    "@babel/plugin-proposal-export-default-from": "^7.2.0",
```
jcger: you forgot the ^

midudev: Thx dude.

PR #516, opened 2019-03-01
[sui-studio] Preserve selects values after reload the pages

carlosvillu:


packages/sui-studio/workbench/src/components/Root/index.js
```diff
@@ -42,7 +42,11 @@ class Root extends React.PureComponent {
     themes: PropTypes.object
   }
 
-  state = {playground, actualContext: 'default', actualStyle: 'default'}
+  state = {
+    playground,
+    actualContext: window.sessionStorage.actualContext || 'default',
```
jcger: Do you mean current instead of actual? although actual might be right, I want to make sure you don't want to use current instead of actual

carlosvillu: I think that actual is fine. but I can change it. What do you prefer ?!

jcger: It's ok then

PR #517, opened 2019-03-05
[sui-studio] Render workbench components inside a shadowDom

carlosvillu:


carlosvillu:
Close this PR because we need support for IE11 and Edge

PR #518, opened 2019-03-05
[sui-studio] Render component inside iFrames and Device selector

carlosvillu:


jcger:
Looks like something is broken, please check `npm run dev molecule/select`  and open an select, it doesn't render as it does in sui-components.now.sh

PR #521, opened 2019-03-06
[sui-ssr] Using the new critical css service and prefetch polyfill

carlosvillu:


packages/sui-ssr/server/dynamicRendering/index.js
```diff
@@ -36,6 +37,17 @@ export default function dynamicRendering(fallback, dynamicsURLS = []) {
             'rel="stylesheet" media="only x" as="style" onload="this.media=\'all\'"'
           )
         )
+        .pipe(
+          replace(
+            HEAD_CLOSING_TAG,
+            `
+        <script>
+        /*! loadCSS. [c]2017 Filament Group, Inc. MIT License */
```
jcger: what about saving this in a const called PREFETCH_POLYFILL and use it? should make this whole stuff more readable...

packages/sui-ssr/server/ssr/index.js
```diff
@@ -48,6 +49,15 @@ export default (req, res, next) => {
         'rel="stylesheet"',
         'rel="stylesheet" media="only x" as="style" onload="this.media=\'all\'"'
       )
+      .replace(
+        HEAD_CLOSING_TAG,
+        `
+        <script>
+        /*! loadCSS. [c]2017 Filament Group, Inc. MIT License */
```
midudev: We might want to extract this to be reused here and with the dynamic rendering.

carlosvillu: Maybe with the Raw loader. But I am not sure if it works in the server config

PR #525, opened 2019-03-09
perf(sui-svg): move to React Hooks instead using class

midudev:
Move sui-svg to hooks in order to improve bundle and parse time.


midudev:
@jcger, the component remains `pure` as I'm wrapping the component with `React.memo`.

PR #548, opened 2019-03-22
fix(sui-studio): move sui-mono as dependency

jcger:
otherwise sui-studio-check-release doesn't work


midudev:
Oooops! Sorry! :D My bad!

PR #552, opened 2019-03-26
feat(sui-bundler): add bundle splitting

kikoruiz:
## Description
Add bundle splitting by browser support (ES5 vs ES6).


carlosvillu:
I really think that we can live with this issue: https://github.com/philipwalton/webpack-esnext-boilerplate/issues/1

packages/sui-bundler/webpack.config.prod.js
```diff
@@ -11,6 +11,8 @@ const webpack = require('webpack')
 const definePlugin = require('./shared/define')
 const babelRules = require('./shared/module-rules-babel')
 
+console.log('new bundler')
```
nucliweb: A console in production, without emojis? 😱

midudev: hahaha still, not production ready. :)

PR #555, opened 2019-03-29
Add "sui-changelog" package

kikoruiz:
# sui-changelog

> CLI to retrieve a changelog from a set of dependencies.

## Description

The main aim of this package is to build a changelog file with all the last changes you get in your last installation within your project. Dependencies are retrieved by its scope in the `node_modules` folder, so you can customize the scopes to get the changelog data from.

Moreover, `sui-changelog` package runs a `npm shrinkwrap` in order to get the differences in the dependencies, and you can optionally run a `phoenix` command if you want to have a clean install before building your changelog.

### Basic usage

Install `node@11` with `npm@6` in your env and run the following CLI command:

```sh
$ npx @s-ui/changelog
```

Or, if you want to save it in your project development dependencies:

```sh
$ npm install @s-ui/changelog --save-dev
$ sui-changelog
```

### Optionally running a phoenix

If you feel more confortable having a clean install of your dependencies before retrieving the changelog, just set the `-p, --phoenix` option:

```sh
$ sui-changelog -p
```

### Adding more scopes

By default, `sui-changelog` works only with `@s-ui` scoped packages, so if you want to add more scopes to your changelog, just add them in your project `package.json` file, the same way as the example below:

```js
{
  "config": {
    "sui-changelog": {
      "scopes": ["@my-awesome-scope", "@another-scope"]
    }
  }
}
```

### Retrieving data from private repositories

If you know you have some private respositories inside your set of dependencies, you should add a GitHub access token as an environment variable (as `GITHUB_TOKEN`) to make it work. Such token has to be added into your `~/.bash_profile` file (or `~/.profile`, or `~/.bashrc`) like the following example:

```cs
export GITHUB_TOKEN="MY_AWESOME_GITHUB_PERSONAL_ACCESS_TOKEN"
```

You can get more information in this link to get the token: https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line. If it's not provided, the changelog data for such package will be ignored.

## Output

After triggering the CLI command `sui-changelog` we'll get new changes in our project. Firstly, current package version (`package.json` file) will be updated. Then, if it's the first time a new file named `CHANGELOG.md` will be added to your project, but if you already got it, a new hunk of changes will be added at the top of your file.

```cs
├── node_modules
    ├── @my-awesome-scope
    │   └── my-awesome-package // Package last changes will be retrieved from.
    ├── @another-scope
    │   └── another-scope-package // Other package last changes will be retrieved from.
    ├── eslint
    └── react
├── package.json // Modified with the new package version.
└── CHANGELOG.md // New or modified with the last changes.
```

### DIFF
--- a/packages/sui-changelog/README.md
+++ b/packages/sui-changelog/README.md
@@ -34,7 +34,7 @@ By default, `sui-changelog` works only with `@s-ui` scoped packages, so if you w
 
 ### Retrieving data from private repositories
 
-If you know you have some private respositories inside your set of dependencies, you might add a GitHub access token in order to make it work. Such token has to be added like the following example:
+If you know you have some private respositories inside your set of dependencies, you should add a GitHub access token in order to make it work. Such token has to be added like the following example:
 
 ```js
 {
@@ -45,3 +45,19 @@ If you know you have some private respositories inside your set of dependencies,
 ```
 
 You can get more information in this link to get the token: https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line. If it's not provided, the changelog data for such package will be ignored.
+
+## Output
+
+After triggering the CLI command `sui-changelog` we'll get new changes in our project. Firstly, current package version (`package.json` file) will be updated. Then, if it's the first time a new file named `CHANGELOG.md` will be added to your project, but if you already got it, a new hunk of changes will be added at the top of your file.
+
+```cs
+├── node_modules
+    ├── @my-awesome-scope
+    │   └── my-awesome-package // Package last changes will be retrieved from.
+    ├── @another-scope
+    │   └── another-scope-package // Other package last changes will be retrieved from.
+    ├── eslint
+    └── react
+├── package.json // Modified with the new package version.
+└── CHANGELOG.md // New or modified with the last changes.
+```
--- a/packages/sui-changelog/README.md
+++ b/packages/sui-changelog/README.md
@@ -6,7 +6,11 @@
 
 The main aim of this package is to build a changelog file with all the last changes you get in your last installation within your project. Dependencies are retrieved by its scope in the `node_modules` folder, so you can customize the scopes to get the changelog data from.
 
+Moreover, `sui-changelog` package runs a `npm shrinkwrap` in order to get the differences in the dependencies, and you can optionally run a `phoenix` command if you want to have a clean install before building your changelog.
+
 ## Usage
+
+### Basic usage
 
 Install `node@11` with `npm@6` in your env and run the following CLI command:
 
@@ -19,6 +23,14 @@ Or, if you want to save it in your project development dependencies:
 ```sh
 $ npm install @s-ui/changelog --save-dev
 $ sui-changelog
+```
+
+### Optionally running a phoenix
+
+If you feel more confortable having a clean install of your dependencies before retrieving the changelog, just set the `-p, --phoenix` option:
+
+```sh
+$ sui-changelog -p
 ```
 
 ### Adding more scopes
--- a/packages/sui-changelog/README.md
+++ b/packages/sui-changelog/README.md
@@ -14,9 +14,10 @@ Install `node@11` with `npm@6` in your env and run the following CLI command:
 $ npx @s-ui/changelog
 ```
 
-Or, if you already have it installed:
+Or, if you want to save it in your project development dependencies:
 
 ```sh
+$ npm install @s-ui/changelog --save-dev
 $ sui-changelog
 ```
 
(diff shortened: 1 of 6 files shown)


packages/sui-changelog/README.md
```diff
@@ -0,0 +1,63 @@
...
+$ sui-changelog
+```
+
+### Adding more scopes
+
+By default, `sui-changelog` works only with `@s-ui` scoped packages, so if you want to add more scopes to your changelog, just add them in your project `package.json` file, the same way as the example below:
+
+```js
+{
+  "sui-changelog": {
```
midudev: In order to follow the @s-ui convention, can we use the `config` key and, inside, use the `sui-changelog` object?

kikoruiz: That's a good point, but I followed the `sui-bundler` example, maybe we should create an issue for moving this change to every `sui` package.

midudev: You're right. @s-ui/bundler is using directly the root 🤔. Could we use the `config` for this package and we will move later the @s-ui/studio? 👏

packages/sui-changelog/README.md
```diff
@@ -0,0 +1,63 @@
...
+```
+
+### Retrieving data from private repositories
+
+If you know you have some private respositories inside your set of dependencies, you should add a GitHub access token in order to make it work. Such token has to be added like the following example:
+
+```js
+{
+  "sui-changelog": {
+    "githubToken": "MY_AWESOME_GITHUB_PERSONAL_ACCESS_TOKEN"
```
jcger: shouldn't this be passed by as environment var or cli argument? We should try to avoid having plain text tokens written in our repos

packages/sui-changelog/bin/index.js
```diff
@@ -0,0 +1,178 @@
...
+      Promise.all(responses).then(repositories => {
+        repositories.forEach((commits, index) => {
+          const {packageScope, packageName, from, to} = modifiedPackages[index]
+          const gitUrl = getRepositoryUrl({packageScope, packageName})
+
+          changelogData.push({h4: `${packageName} ${to}`})
+
+          // Maybe there is no GitHub URL for this package or it's
+          // invalid so we cannot get its changelog data.
+          // So that we print an special message.
```
jcger: **a** special message

packages/sui-changelog/bin/index.js
```diff
@@ -0,0 +1,178 @@
...
+
+  log('MODIFIED PACKAGES:')
+  hunks.forEach(({lines}) => {
+    // Get modified packages by filtering lines with git removal syntax from
+    // diff.
+    const linesWithModifiedPackages = lines.filter(line =>
+      line.match(oldVersionRegExp)
+    )
+    linesWithModifiedPackages.forEach(oldVersionLine => {
+      const oldVersionIndex = lines.indexOf(oldVersionLine)
```
jcger: you could extract this kind of callbacks with a descriptive name in order to make the code more readable

packages/sui-changelog/README.md
kikoruiz: Yep, I'm doing it right for this package 😉

packages/sui-changelog/bin/index.js
```diff
@@ -0,0 +1,178 @@
...
+const log = console.log
+const error = console.error
+const exit = process.exit
+const date = new Date()
+const {
+  DEFAULT_SCOPES,
+  PACKAGE_FILES,
+  MAX_BUFFER,
+  version,
+  scopes: customScopes = [],
```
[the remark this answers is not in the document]
kikoruiz: I just wanted to use a `join` method in a relaxed way 😂

packages/sui-changelog/README.md
kikoruiz: @jcger and @carlosvillu You're right, this is not a good way of working with secrets, but at the end of the day, with any of these methods, we'll have the JARVIS token written in all our private repos. Just to make it clear, with your suggestion, this bin script will be called from a NPM script like this:
```json
{
  "scripts": {
    "changelog": "sui-changelog GITHUB_TOKEN=1234abcd"
  }
}
```

carlosvillu: I mean, you have to call without define in the command any env var. And later inside your code check is this vars exist.
https://github.com/SUI-Components/sui/blob/master/packages/sui-ssr/bin/sui-ssr-release.js#L46-L50

packages/sui-changelog/README.md
midudev: Thanks kiko! 🙇‍♂️

packages/sui-changelog/README.md
kikoruiz: Clear! It's already applied the changes you suggest 😉

packages/sui-changelog/bin/index.js
kikoruiz: Done!

PR #559, opened 2019-04-03
chore(META): use manual sui-mono release and remove npm scripts

midudev:
Remove npm scripts for releasing packages.
Use manual release command in travis.


midudev:
Julian told me that could be a good idea to keep the scripts but do an "echo" to inform about the new behaviour. So I will iterate the solution to add it. :)

PR #561, opened 2019-04-03
Feature/update sui bundler widget version

midudev:
Update @s-ui/bundler version on widget-embedder.

### DIFF
--- a/packages/sui-widget-embedder/src-react/render.js
+++ b/packages/sui-widget-embedder/src-react/render.js
@@ -1,3 +1,5 @@
+/* eslint-disable */
+
 import React from 'react'
 import ReactDOM from 'react-dom'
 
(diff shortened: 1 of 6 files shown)


packages/sui-widget-embedder/src-react/render.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable */
```
carlosvillu: 👀

midudev: It's in purpose! I had problems while linking the package with errors on the compiled one. This will ensure the "magic" comment of eslint is kept after preparing the react files so the sui-bundler doesn't give problems. Maybe that happens only on linking the sui-widget-embedder, I could remove them and try again with the published package.

PR #566, opened 2019-04-09
Feature/sui test build docker image

carlosvillu:


packages/sui-test/bin/sui-test-e2e.js
```diff
@@ -6,7 +6,7 @@ const program = require('commander')
 const fs = require('fs')
 const {getSpawnPromise, showError} = require('@s-ui/helpers/cli')
 const {resolveLazyNPMBin} = require('@s-ui/helpers/packages')
-const CYPRESS_VERSION = '3.0.3'
+const CYPRESS_VERSION = '3.0.3' // Change this version in the Dockerfile too!!!!
```
kikoruiz: Don't like too much this duplicity of versions...

carlosvillu: I could move this to the package.json. And read it from there.

PR #579, opened 2019-04-29
feat(sui-analytics)

jcger:
Segments HTTP API abstraction as defined by https://segment.com/docs/sources/server/http/

TODO (will be done in new PR's)
- [ ] fill README.md file
- [ ] should we send our package.json version within every call for easier debugging?
- [ ] handle events fired before lib loaded
- [ ] handle retries
- [ ] handle response error. Segment’s API will respond with 400 Bad Request if request size is too big
- [ ] check if common fields are correct https://segment.com/docs/spec/common/
- [ ] clarify if we should use `screen` or `page` event


packages/sui-analytics/src/events/index.js
```diff
@@ -0,0 +1,3 @@
+export {default as identify} from './identify'
+export {default as track} from './track'
+export {default as screen} from './screen'
```
kikoruiz: Not sure if screen is the most appropriated name for page views... 🤔

packages/sui-analytics/src/events/screen.js
```diff
@@ -0,0 +1,13 @@
...
+import getAnonymousId from '../anonymousUser'
+
+export default data => ({
+  originalTimestamp: new Date().toISOString(),
+  anonymousId: getAnonymousId(),
+  category: data.category,
+  context,
+  name: data.name,
+  properties: {...data.properties},
+  type: 'screen',
```
kikoruiz: Is `type` property needed?

kikoruiz:
Did you consider using `batch` method (https://segment.com/docs/sources/server/http/#batch)? In some cases it could help us to save some extra requests.

packages/sui-analytics/src/events/index.js
jcger: me either, one pro of calling it screen is that we will have a common language with data and apps

packages/sui-analytics/src/events/screen.js
jcger: looks like it doesn't, will fix it

jcger:
> Did you consider using `batch` method (https://segment.com/docs/sources/server/http/#batch)? In some cases it could help us to save some extra requests.

Yes, will use it after collecting possible events tracked before the lib was loaded

packages/sui-analytics/.babelrc
```diff
@@ -0,0 +1,3 @@
+{
```
carlosvillu: I think that you can remove this file

packages/sui-analytics/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "main": "index.js",
+  "scripts": {
+    "lib": "npx rimraf ./lib && npx mkdirp ./lib && babel --presets sui ./src --out-dir ./lib",
+    "test": "echo \"Error: no test specified\" && exit 1"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
+  "dependencies": {
+    "@s-ui/js": "1.26.0",
```
carlosvillu: 👀

packages/sui-analytics/src/anonymousUser/cookies.js
```diff
@@ -0,0 +1,10 @@
+import cookie from '@s-ui/js/lib/cookie'
```
carlosvillu: This file is an abstraction already. You do not need another layer.

packages/sui-analytics/src/anonymousUser/index.js
```diff
@@ -0,0 +1,22 @@
+import cookieStrategy from './cookies'
+import localStorageStrategy from './localStorage'
+import uuid from 'uuid/v4'
+
+const KEY = 'sui-analytics-anonymous-user-id'
+
+const getAnonymousUserId = function(key) {
+  return (
+    cookieStrategy.getItem(key) || localStorageStrategy.getItem(key) || uuid()
```
carlosvillu: Maybe a try/catch strategy here could make sense. There are scenarios like Safari InPrivate mode where try to access to LS will throw an exception.

packages/sui-analytics/src/events/index.js
carlosvillu: Use the same language than Data and Apps I think that will be the best.

kikoruiz: You're right, in Motor Data&Insights are already using "screen" concept also for page views 👍

PR #584, opened 2019-05-08
Added htmr package

nucliweb:
## Description

Added a new package `htmr` to have the option to convert html strings to React element.
`No more dangerouslySetInnerHTML`

## More info

👉🏼 [htmr](https://github.com/pveyes/htmr)

## Example

```
const HtmlContent = (html) => {
  return (
    htmlStringToReactElement(html)
  )
}
```


rmoralp:
Could you tell an use case where do you need this?

packages/sui-js/package.json
```diff
@@ -28,7 +28,8 @@
   },
   "devDependencies": {
     "@babel/cli": "7",
-    "babel-preset-sui": "3"
+    "babel-preset-sui": "3",
+    "htmr": "^0.7.0"
```
rmoralp: May we remove the caret?

nucliweb: Removed!

PR #592, opened 2019-05-16
[sui-ssr] Add support for multisite.

kikoruiz:
## Description
In some cases our SPA can be multisite, it means it works with the same code but it's used by different domains or hostnames. This feature keeps working the previous behaviour (one single public folder) by default but it can be enabled by setting a mapping with an association of hostname (key) and public folder (value). See a simple example below:
```json
{
  "multiSite": {
    "my-motorcycles.com": "public-motorcycles",
    "my-trucks.com": "public-trucks",
    "default": "public-cars"
  }
}
```

**Important! At least a `default` value is mandatory to enable this feature.**

Once this set is done, if you want to test your server in `localhost` you must run it setting the custom header `'X-Serve-Site'` to the request. If you're a Google Chrome user, you can achieve it by installing the extension [ModHeader](https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj).


packages/sui-ssr/server/dynamicRendering/index.js
```diff
@@ -46,7 +49,7 @@ export default function dynamicRendering(fallback, dynamicsURLS = []) {
           )
         )
     } else {
-      indexHTMLStream = fs.createReadStream(INDEX_HTML_PATH)
+      indexHTMLStream = fs.createReadStream(indexHtmlPath(req))
```
naxhh: not related to the PR change but, are we reading the file on each request? wouldn't that hurt performance wise?

carlosvillu: We are doing that because after closing the pipe, the next client will get a closed pipe. But you are right. We must move this from a Pipe to an inmemory var.

packages/sui-ssr/README.md
```diff
@@ -192,10 +192,25 @@ Configs accepted:
...
+  ```json
+  {
+    "multiSite": {
+      "my-motorcycles.com": "public-motorcycles",
+      "my-trucks.com": "public-trucks",
+      "default": "public-cars"
+    }
+  }
+  ```
+  Once this set is done, if you want to test your server in `localhost` you must run it setting the custom header `'X-Serve-Site'` to the request. If you're a Google Chrome user, you can achieve it by installing the extension [ModHeader](https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj).
```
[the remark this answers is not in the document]
kikoruiz: In fact, if we access localhost running the server, it will load the `default` version, but this guide is for, optionally, accessing another site in runtime.

packages/sui-ssr/server/dynamicRendering/index.js
kikoruiz: Do you mean `indexHTMLStream` or `indexHtmlPath` variable?

carlosvillu: `indexHTMLStream `

PR #597, opened 2019-05-24
feat(sui-studio): adds a mobile device just to make it look cooler

davecarter:
### Summary
Adds a mobile device graphic to give a nicer context to the rendered component in mobile viewports.
Here is an attached screen caption

Please review:

mobile

### DIFF
--- a/packages/sui-studio/workbench/src/components/Root/index.js
+++ b/packages/sui-studio/workbench/src/components/Root/index.js
@@ -1,9 +1,9 @@
 /* eslint import/no-webpack-loader-syntax:0 */
-import React from "react";
-import PropTypes from "prop-types";
+import React from 'react'
+import PropTypes from 'prop-types'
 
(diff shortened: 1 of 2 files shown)


packages/sui-studio/workbench/src/components/Root/index.js
```diff
@@ -1,24 +1,24 @@
 /* eslint import/no-webpack-loader-syntax:0 */
-import React from 'react'
-import PropTypes from 'prop-types'
+import React from "react";
```
midudev: It seems you're not using the correct linter as simple quotes are the one we're using to format our code. Recheck it!

midudev: Aslo there's some semicolons in the file :O

davecarter: not anymore.. 😅

PR #616, opened 2019-06-28
feat(sui-i18n): (backwards compatible) add an option to allow query p…

arnau-rius:
## Description
Add an option to the url method in order to allow ? and = signs (query paramters)  (backwards compatible).

## Related Issue
Before this PR, if you add a query parameter in the url, the chars: `?` and `=` are removed from the url

## Example
Before: https://milanuncios.com/mis-anuncios/?recargar=1 is converted to https://milanuncios.com/mis-anuncios/recargar1
After: https://milanuncios.com/mis-anuncios/?recargar=1 is converted to https://milanuncios.com/mis-anuncios/?recargar=1

### DIFF
--- a/packages/sui-i18n/src/i18n.js
+++ b/packages/sui-i18n/src/i18n.js
@@ -96,10 +96,10 @@ export default class Rosetta {
     throw new Error(`Invalid type '${type}' passed to i18n.f`)
   }
 
-  url(urlPattern) {
+  url(urlPattern, allowQueryParams) {
     return urlPattern
       .split('/')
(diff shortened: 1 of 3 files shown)


packages/sui-i18n/src/i18n.js
```diff
@@ -96,10 +96,10 @@ export default class Rosetta {
     throw new Error(`Invalid type '${type}' passed to i18n.f`)
   }
 
-  url(urlPattern) {
+  url(urlPattern, allowQueryParams) {
```
carlosvillu: Update the Readme.

PR #618, opened 2019-07-01
Feat/move to stylelint

midudev:
THIS IS A BREAKING CHANGE!

Move to stylelint and stop using sasslint.
Add prettier to our .scss files.
Add info about how to configure it.


packages/sui-lint/stylelint.config.js
```diff
@@ -0,0 +1,65 @@
...
+    'block-no-empty': true,
+    'color-no-invalid-hex': true,
+    'comment-no-empty': true,
+    'declaration-block-no-duplicate-properties': [
+      true,
+      {
+        ignore: ['consecutive-duplicates-with-different-values'] // useful for fallback properties
+      }
+    ],
+    'declaration-block-no-shorthand-property-overrides': true,
```
davecarter: Just wondering if all these `stylelint` rules and the previous ones defined by `sass-lint' are going to have quite differences between them..

midudev: *Tried* to keep exactly the same rules plus added some recommended ones. Some are just breaking, like the zero leading, as the stylelint-prettier is enforcing it (and is not configurable).

PR #620, opened 2019-07-02
feat(sui-js): arrayToCommaQueryString method

AgonisticKatai:
## Description
Method for transforming a parameter object into a query-string. If one of the parameters contains an array, it will construct it with a comma separation.

> Needs to upgrade `qs` package from `6.6.0` to `6.7.0`

## Example
```javascript
const params = {userId: 1, adId: 2, products: [3, 4, 5]}
console.log(arrayToCommaQueryString(params)) // 'a=1&b=2&c=3,4,5'
```

### DIFF
--- a/packages/sui-js/README.md
+++ b/packages/sui-js/README.md
@@ -96,9 +96,9 @@ import {parseQueryString} from '@s-ui/js/lib/string'
 console.log(parseQueryString('?targetPage=pta')) // {targetPage: "pta"}
 
 
-import {arrayToCommaQueryString} from '@s-ui/js/lib/string'
+import {fromArrayToCommaQueryString} from '@s-ui/js/lib/string'
 
-console.log(arrayToCommaQueryString({userId: 1, adId: 2, products: [3, 4, 5]})) // 'userId=1&adId=2&products=3,4,5'
+console.log(fromArrayToCommaQueryString({userId: 1, adId: 2, products: [3, 4, 5]})) // 'userId=1&adId=2&products=3,4,5'
 
 
 import {htmlStringToReactElement} from '@s-ui/js/lib/string'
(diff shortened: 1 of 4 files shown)


packages/sui-js/src/string/index.js
```diff
@@ -1,9 +1,13 @@
-import {parse} from 'qs'
+import {parse, stringify} from 'qs'
 
 export const parseQueryString = query => parse(query, {ignoreQueryPrefix: true})
+export const arrayToCommaQueryString = query =>
+  unescape(stringify(query, {arrayFormat: 'comma'}))
```
rmoralp: unescape 'fn' is deprecated
why not import `stringify` and send the options from outside? I think that this is unnecessary

packages/sui-js/README.md
```diff
@@ -96,6 +96,11 @@ import {parseQueryString} from '@s-ui/js/lib/string'
 console.log(parseQueryString('?targetPage=pta')) // {targetPage: "pta"}
 
 
+import {arrayToCommaQueryString} from '@s-ui/js/lib/string'
+
+console.log(arrayToCommaQueryString({userId: 1, adId: 2, products: [3, 4, 5]})) // 'a=1&b=2&c=3,4,5'
```
kikoruiz: I would expect `userId=1&adId=2&products=3,4,5` to be the output.

AgonisticKatai: Copy and paste fail... 🤓

packages/sui-js/src/string/index.js
AgonisticKatai: changed it to `decodeUriComponent()`

packages/sui-js/README.md
AgonisticKatai: FIXED!

PR #623, opened 2019-07-04
feat(sui-domain): accept multiple params for usecase.execute method

jcger:
## Description
Allows passing by multiple args to the useCase's execute method

## Related Issue
I would like to follow [Segment's track spec](https://segment.com/docs/spec/track/) in order to have the same API everywhere and be able to execute something like:

```js
domain.get('myTrackingEventUseCase').execute('Event Name', {eventProp1: ''})
```

### DIFF
--- a/packages/sui-domain/src/EntryPointFactory.js
+++ b/packages/sui-domain/src/EntryPointFactory.js
@@ -40,10 +40,12 @@ export default ({useCases, config}) =>
       return loader === undefined
         ? new NotImplementedUseCase(key)
         : {
-            execute: params => {
+            execute: (...params) => {
               // load async the factory, execute use case and return the promise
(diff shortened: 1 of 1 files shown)


packages/sui-domain/src/EntryPointFactory.js
```diff
@@ -40,10 +40,12 @@ export default ({useCases, config}) =>
       return loader === undefined
         ? new NotImplementedUseCase(key)
         : {
-            execute: params => {
+            execute: (...params) => {
```
carlosvillu: I don't think that this could be a good approach. We must continue using named params over positional params.
And do whatever you want in your use case

jcger: why?

PR #624, opened 2019-07-08
feat(sui-bundler): add alias to keep same experiment context on linked packages

desko27:
## ✍️ Description
The **Experiment Context** feature [which was introduced in this PR](https://github.com/SUI-Components/schibsted-spain-components/pull/94) and resides in the `@s-ui/abtesting-optimizely-x` package, needs an alias on sui-bundler in order to work with linked packages, so the `createContext` method is ran from a single place: the `node_modules` directory from the project in which sui-bundler is running.

This is the very same problem that is being solved for `@s-ui/react-context` package by adding an alias as well:

https://github.com/SUI-Components/sui/blob/3fd5a279e43e9aed038b48da41b3a4d5774bfe72/packages/sui-bundler/webpack.config.dev.js#L28-L30

### DIFF
--- a/packages/sui-bundler/README.md
+++ b/packages/sui-bundler/README.md
@@ -180,7 +180,6 @@ This tool works with zero configuration out the box but you could use some confi
 }
 ```
 
-**advice:** the alias option will not work for dev environment
 > The URL to the CDN **MUST** end with a slash `/`
 
 ## OnlyHash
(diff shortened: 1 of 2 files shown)


desko27:
> We're going to rethink all of this about alias as this clearly doesn't seem to be agnostic to our stack. We might could extract this to a config to be used for each app instead having it here inlined.

@midudev I totally agree. So I understand that we're considering this a temporary solution along with the already existing aliases.

desko27:
@midudev @carlosvillu thank you so much for the feedback.

I pushed some changes in order to apply the suggestion from @carlosvillu, so I removed the hardcoded alias while enabling the host project to use its preferred list of aliases in development environment.

I tested this and it's working nice:
```
  "sui-bundler": {
    "alias": {
      "@s-ui/abtesting-optimizely-x": "./node_modules/@s-ui/abtesting-optimizely-x"
    },

    ...
```

PR #628, opened 2019-07-16
feat(sui-analytics): add adobe marketing cloud id tracking

jcger:
We have a common orgId but I'm not sure if we should hardcode it here, therefore, I added a way to add this id but we have to set it as soon as possible

### DIFF
--- a/packages/sui-analytics/src/index.js
+++ b/packages/sui-analytics/src/index.js
@@ -1,3 +1,5 @@
+let adobeOrgId
+
 export default {
   identify: (...args) => {
     window.analytics.identify(...args)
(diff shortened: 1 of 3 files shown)


packages/sui-analytics/src/index.js
```diff
@@ -12,7 +14,17 @@ export default {
         anonymousId: window.analytics.user().anonymousId(),
         userId: window.analytics.user().id()
       }
+      if (window && window.Visitor && adobeOrgId) {
+        options['Adobe Analytics'] = {
+          marketingCloudVisitorId: window.Visitor.getInstance(
```
carlosvillu: I think that maybe this is too much business-related to being here. Maybe we can develop the concept of middleware inside of sui-analytic

PR #636, opened 2019-07-26
feat(sui-js): upgrade bowser version and add method uaParser factory

giodelabarrera:
Changes in `sui-js`
- Upgrade bowser dependency version to `2.5.2` maintaining compatibility of `stats` method.
- Add factory method to call only once `uaParser`.
- Add method `satisfies` to validate browsers.
- Add method `parse` to get full info.


kikoruiz:
It could be a good idea to perform some tests in this tool in order to ensure the complete compatibility.

PR #648, opened 2019-08-21
Feat/sui bundler markdown url

davecarter:
## Description
- [x] In order to minify url mismatches loading the `manifestURL` we need to simplify the RegEx.
- [x] When getting the Manifest we are moving to this object instead of a single string without breaking backwards compatibility:
```
  "sui-bundler": {
    "externals-manifest": {
         "preproduction": "https://scm-frontend-pre.MY-SITE.com/",
         "production": "https://scm-frontend-pro.MY-SITE.com/"
    }
}
```
Please review


packages/sui-bundler/loaders/ExternalsManifestLoader.js
```diff
@@ -34,13 +34,16 @@ async function externalsManifestLoader(source) {
...
+      typeof manifestURL === 'string'
+        ? manifestURL
+        : manifestURL[process.env.NODE_ENV]
+    )
     const entries = Object.entries(manifest)
     const nextSource = entries.reduce((acc, entry) => {
       const [dest, src] = entry
-      const regex = new RegExp(`${dirname}${dest}`, 'g')
-      return acc.replace(regex, `${dirname}${src}`)
+      const regex = new RegExp(`${dest}`, 'g')
```
midudev: The template string doesn't make sense anymore. Neither on the L46.

davecarter: I thought that the RegEx needs to have an string...

midudev: And what's the `dest` and `src` type?

davecarter: ¯\_(ツ)_/¯  .. oks 😅

PR #652, opened 2019-09-02
feat(sui-decorators): allow to disable the cache.

tureey:
## Description
The decorator applies the cache always. In some cases we may need to disable this in certain environment such as testing. For this purpose, add a `__SUI_CACHE_DISABLED__` variable in global scope.

## Example
```
// For client side
window.__SUI_CACHE_DISABLED__ = true

// Server side
global.__SUI_CACHE_DISABLED__ = true
```

### DIFF
--- a/packages/sui-decorators/src/decorators/cache/index.js
+++ b/packages/sui-decorators/src/decorators/cache/index.js
@@ -37,7 +37,7 @@ const _cache = ({
       (typeof window !== 'undefined' && window.__SUI_CACHE_DISABLED__) ||
       (typeof global !== 'undefined' && global.__SUI_CACHE_DISABLED__)
     ) {
-      return original
+      return original.apply(instance, args)
     }
 
     const key = `${target.constructor.name}::${fnName}::${createHash(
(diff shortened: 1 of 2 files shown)


packages/sui-decorators/src/decorators/cache/index.js
```diff
@@ -33,6 +33,13 @@ const _cache = ({
   const cache = caches[fnName]
 
   return (...args) => {
+    if (
+      (typeof window !== 'undefined' && window.__SUI_CACHE_DISABLED__) ||
+      global.__SUI_CACHE_DISABLED__
```
jcger: shouldn't you check if `global` exists?

packages/sui-decorators/src/decorators/cache/index.js
```diff
@@ -33,6 +33,13 @@ const _cache = ({
...
+    ) {
+      return original
```
carlosvillu: I think that you should execute the original function.

```
original.apply(instance, args)
```

packages/sui-decorators/src/decorators/cache/index.js
tureey: Oh, I thought it would exist if `window` didn't !

packages/sui-decorators/src/decorators/cache/index.js
tureey: Yep, I quick read the function and I thought that `args` where the decorator `args`. My bad

PR #654, opened 2019-09-16
Feat/add spec tab

midudev:
- Add UX Definition tab
- Add custom snarkdown implementation with table support.
- Move from class to function several components.
- Change template for component creation to use function by default.

image

If the file doesn't exist, it removes the button:
image


turolopezsanabria:
👏 👍 qué preciosidad

PR #656, opened 2019-09-26
feat(sui-studio): generate componente in functional syntax instead of…

arnau-rius:
## Description
- `sui-studio generate alpha beta` will generate components that will be based in *functional syntax* instead of *class based syntax*.


midudev:
Studio is already generating function based components.

PR #657, opened 2019-09-27
feat(sui-lint): bump dependencies

midudev:
Bump dependencies and add lots of new changes to linter.

Major changes 👀 ⬇️ 
- ❗️ Drop support for node <8.9.0.

New errors ❌ :
- js: Using target="_blank" without rel="noopener noreferrer" is a security risk.
- js: `object["string"]` is better written in dot notation `object.string`.
- ⚛️Missing "key" prop for element in array
- ⚛️ Prefer fragment shorthand (<>) over React.Fragment
- ⚛️ Avoid unnecessary curly braces for string. url={'#'} ➡️  url="#"
- ⚛️  More strict usage of hooks. Now they can't be used on the root level of the file (outside the component).

New warnings ⚠️ :
- Handler functions prop key must begin with 'handle'.
- Prop key for events must begin with 'on'.
- ⚛️ Deprecation warning for unsafe life cycle methods.


jcger:
about `js: object["string"] is better written in dot notation object.string.`
what if the string has whitespaces in between? Sometimes there is no way to avoid it as they are set or needed by the API which we are not owners from

about `Prefer fragment shorthand (<>) over React.Fragment`
is there anything wrong about using React.Fragment? I'd like to avoid throwing errors for something that aren't errors nor wrong by itself

midudev:
Hi @jcger, about the first one. If the string has whitespaces, you could safely use the `object["  string"]` or use a variable (which would be a better idea) `object[variableWithTheString]`. Both should work without problems.

About the later. It's not anything wrong with React.Fragment. It's a way to avoid having three different ways of doing the same: `<>`, `<Fragment>`, `<React.Fragment>` and embrace only one of them for consistency among all the codebase.

I could overwrite to be a warning for now. But we're throwing errors for things that are not errors itself but to keep consistency (like spacing or usage of brackets).

PR #663, opened 2019-10-01
Fix/js linter problems

midudev:
Use latest sui-lint to fix all Javascript linter errors on the project.

### DIFF
--- a/packages/sui-changelog/package.json
+++ b/packages/sui-changelog/package.json
@@ -32,7 +32,7 @@
   },
   "devDependencies": {
     "@s-ui/lint": "3",
-    "@s-ui/mono": "1.43.0",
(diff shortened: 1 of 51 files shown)


packages/sui-changelog/package.json
```diff
@@ -32,16 +32,11 @@
   },
   "devDependencies": {
     "@s-ui/lint": "3",
-    "@s-ui/mono": "1",
+    "@s-ui/mono": "1.43.0",
```
jcger: really?

midudev: That's very weird. :O I didn't add this line at all.

midudev: Fixed ✔️

PR #676, opened 2019-10-15
fix(sui-studio): fix interops for default exports

carlosvillu:
image

---
This change is [Reviewable](https://reviewable.io/reviews/sui-components/sui/676)


midudev:
Instead doing this, we're going to do a rollback over the raw-loader to also fix problems on normal mode and generating the statics: https://github.com/SUI-Components/sui/pull/677

PR #679, opened 2019-10-21
feat(sui-ssr): added configuration to set the server content-type

ferransimon:
## Description
Added a new configuration property `serverContentType` that if is set it will set the current Content-type of the server.

### DIFF
--- a/packages/sui-ssr/server/ssr/index.js
+++ b/packages/sui-ssr/server/ssr/index.js
@@ -37,7 +37,7 @@ const useLegacyContext =
     : true
 
 const setContentType = res => {
-  if (ssrConfig.serverContentType !== 'undefined') {
+  if (ssrConfig.serverContentType) {
     res.set('Content-Type', ssrConfig.serverContentType)
   } else {
     res.type('html')
--- a/packages/sui-ssr/server/ssr/index.js
+++ b/packages/sui-ssr/server/ssr/index.js
@@ -36,8 +36,16 @@ const useLegacyContext =
     ? ssrConfig.useLegacyContext
     : true
 
+const setContentType = res => {
+  if (ssrConfig.serverContentType !== 'undefined') {
+    res.set('Content-Type', ssrConfig.serverContentType)
+  } else {
+    res.type('html')
+  }
+}
+
 const initialFlush = res => {
-  res.type('html')
+  setContentType(res)
   res.flush()
 }
 
--- a/packages/sui-ssr/README.md
+++ b/packages/sui-ssr/README.md
@@ -213,6 +213,8 @@ Configs accepted:
   ```
   Once this set is done, if you want to test your server in `localhost` you must run it setting the custom header `'X-Serve-Site'` (with the value of your desired site) to the request. If you're a Google Chrome user, you can achieve it by installing the extension [ModHeader](https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj).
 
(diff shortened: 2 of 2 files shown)


packages/sui-ssr/server/ssr/index.js
```diff
@@ -36,8 +36,16 @@ const useLegacyContext =
     ? ssrConfig.useLegacyContext
     : true
 
+const setContentType = res => {
```
carlosvillu: Maybe a Hook is a better place to set up the content-type of your req.

packages/sui-ssr/server/ssr/index.js
```diff
@@ -36,8 +36,16 @@ const useLegacyContext =
...
+  if (ssrConfig.serverContentType !== 'undefined') {
```
carlosvillu: Update Readme.

packages/sui-ssr/server/ssr/index.js
ferransimon: We've tried with a hook but seems that this is executed after any hook and so the Content-Type always gets override

packages/sui-ssr/server/ssr/index.js
ferransimon: done!

packages/sui-ssr/server/ssr/index.js
```diff
@@ -36,8 +36,16 @@ const useLegacyContext =
...
+    res.set('Content-Type', ssrConfig.serverContentType)
+  } else {
+    res.type('html')
```
midudev: A better approach would be to add a defaults values to config in order to avoid this kind of branches in the middle of the code.

ferransimon: well as you can see one branch is using res.type and the other res.set and in order to keep backward compatibility, we've decided to go with this approach

packages/sui-ssr/server/ssr/index.js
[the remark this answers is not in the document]
ferransimon: Thanks for pointing this out, didn't see it. fixed!

packages/sui-ssr/server/ssr/index.js
midudev: Indeed. What I see this branch could be easily avoided by using a default value to the `ssrConfig.serverContentType`. Why? If not, we're going to polute the whole code with branching everytime a value is missing.

We could add on the `config.js` the default value.

``` //config.js
const DEFAULT_VALUES = {
  serverContentType: 'html'
}

export default { ...DEFAULT_VALUES, ...ssrConfig }
```

Later, instead the if you could simply:
```
res.type(ssrConfig.serverContentType)
```

As mentioned on the documentation, type will set the Content-Type header when the string passed adds a `/`: https://expressjs.com/es/api.html#res.type

midudev: Said that, I think is fair to merge this, as it's doing what's mean to do, and we could add later ;)

ferransimon: I do not see how we can set the charset encoding using the approach you've mentioned. To add more context to this PR our problem is that we are using a charset encoding that is not `utf-8` and seems that using only `res.type('html')` by defaults sets the charset encoding to `utf-8` that's why we decided to use the res.set where you can explicitly set the charset encoding.

midudev: Easy: `res.type('text/html;charset=windows-1252')`

ferransimon: oh nice, I didn't understood what the documentation means to say with `If type contains the “/” character, then it sets the Content-Type to type.`

PR #692, opened 2019-11-04
Feat/sui svg using atom icon

midudev:
🔨 This is a **BREAKING CHANGE** as the SVG ICON will be wrapped with [SuiAtomIcon](https://sui-components.now.sh/workbench/atom/icon/demo). That means you should follow the rules provided by the [UX Definition](https://sui-components.now.sh/workbench/atom/icon/documentation/uxdef).

- @s-ui/svg now wraps icon with AtomIcon in order to follow strict guidelines for icons across all verticals.
- Remove @composer dependencies for demo in order to create a custom one.
- Install @s-ui/react-atom-icon dependency on iconset if not installed yet.
- Refactor code to improve readability overall when working with promises.
- Add clean of `lib` folder here instead of every `iconset` package.
- Use the most of `fs-extra`and use their Promises.

Image:
image

Example of output icon:
```
import React, { memo } from "react";
import AtomIcon from "@s-ui/react-atom-icon";
export default memo(function (props) {
  return React.createElement(AtomIcon, props, React.createElement("svg", {
    viewBox: "0 0 64 64"
  }, React.createElement("path", {
    d: "M46.586 58.869a1.998 1.998 0 01-2.825.003l-.003-.003-25.456-25.456a1.998 1.998 0 01-.002-2.826l.002-.002L43.758 5.129a2 2 0 112.828 2.828L22.544 31.999l24.042 24.042c.781.78.782 2.045.002 2.826l-.002.002z"
  })));
});
```


packages/sui-svg/bin/sui-svg-build.js
```diff
@@ -26,64 +32,64 @@ program
...
-  }
-}
+const getAllSrcSvgFiles = () => fg([`${SVG_FOLDER}/**/*.svg`])
 
-function ensureDirectoryExistence(dirname) {
-  if (directoryExists(dirname)) {
-    return true
-  }
-  fs.mkdirSync(dirname)
+const transfomrSvgToReact = svg =>
```
giodelabarrera: `transfomrSvgToReact` -> `transformSvgToReact`

packages/sui-svg/package.json
```diff
@@ -10,16 +10,16 @@
...
+    "@babel/core": "7.6.4",
     "@s-ui/bundler": "5",
     "@s-ui/component-peer-dependencies": "latest",
     "@s-ui/helpers": "1",
+    "@s-ui/react-atom-icon": "1",
     "@svgr/core": "3.1.0",
-    "commander": "2.19.0",
-    "fast-glob": "2.2.4",
-    "lodash.camelcase": "4.3.0",
-    "styled-components": "4.1.1"
```
giodelabarrera: 👋🏽😭

midudev: hahaha It was only used internally by composer. :P Nothing to see here! Keep walking! 🚶

PR #697, opened 2019-11-12
feat(sui-studio-utils): adds compatibility with inlineError

davecarter:
## Description
Currently, Domain useCases using the [@inlineError](https://github.com/SUI-Components/sui/blob/master/packages/sui-decorators/src/decorators/error.js#L6) decorator are not supported by `DomainBuilder` to override responses when Demoing components.

This PR adds compatibility so we'll be able to use both: Domain useCases using the pattern `[error, response]` and `response` only.

- [x] Refactored nested ternaries to improve readability. 
Thanks to @midudev and @zecafa for your help 😍 

## Example
Let's say I want to build a Demo for some component using the current logged `userId`. We can simulate a `get_current_user_use_case` response via DomainBuilder:

```javascript
domain: DomainBuilder.extend({domain: new Milanuncios()})
   .for({useCase: 'get_current_user_use_case'})
   .respondWith({
        success: {id: 179587252}
})
```
So far, DomainBuilder was not taking into account possible useCase decorated responses with `@inlineError` so we were unable to use this feature. This PR fixes this scenario.

### DIFF
--- a/packages/sui-studio-utils/src/domain-builder/index.js
+++ b/packages/sui-studio-utils/src/domain-builder/index.js
@@ -50,34 +50,27 @@ export default class DomainBuilder {
 
   build({inlineError}) {
     const self = this
+    const exeUseCase = useCase => params => {
+      const {fail, success} = self._useCases[useCase]
+
+      const data = typeof success === 'function' ? success(params) : success
+
+      const responseParams =
+        data !== undefined ? {err: null, data} : {err: fail, data: null}
+
+      const createResponse = ({err, data}) => {
+        const {resolve, reject} = Promise
+        if (inlineError) return resolve([err, data])
+
+        return err ? reject(err) : resolve(data)
+      }
+      return createResponse(responseParams)
+    }
+
     return {
-      get: useCase => {
-        if (useCase === 'config') {
-          return this._config
-        }
-        return self._useCases[useCase]
-          ? {
-              execute: params => {
-                const successResponse = self._useCases[useCase].success
-                return self._useCases[useCase].success !== undefined
-                  ? Promise.resolve(
-                      typeof successResponse === 'function'
-                        ? inlineError
-                          ? [null, successResponse(params)]
-                          : successResponse(params)
-                        : inlineError
-                        ? [null, successResponse]
-                        : successResponse
-                    )
-                  : Promise.resolve(
-                      inlineError
-                        ? [self._useCases[useCase].fail, null]
-                        : self._useCases[useCase].fail
-                    )
-              }
-            }
-          : self._domain.get(useCase)
-      },
+      get: useCase => ({
+        execute: exeUseCase(useCase)
+      }),
       _map: Object.assign(self._domain._map || {}, self._useCases),
       useCases: Object.assign(self._domain.useCases || {}, self._useCases)
     }
--- a/packages/sui-studio-utils/src/domain-builder/index.js
+++ b/packages/sui-studio-utils/src/domain-builder/index.js
@@ -1,4 +1,3 @@
-/* eslint-disable no-nested-ternary */
 export default class DomainBuilder {
   static extend({domain = {}}) {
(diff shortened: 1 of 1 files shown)


packages/sui-studio-utils/src/domain-builder/index.js
```diff
@@ -1,15 +1,18 @@
+/* eslint-disable no-nested-ternary */
```
kikoruiz: Why? 🙄

midudev: Because nesting ternaries are a warning.

packages/sui-studio-utils/src/domain-builder/index.js
```diff
@@ -1,15 +1,18 @@
...
 export default class DomainBuilder {
   static extend({domain = {}}) {
     return new DomainBuilder({domain})
   }
 
+  // @ts-ignore
```
midudev: ¿?

davecarter: The linter is complaining about this...
deleted

packages/sui-studio-utils/src/domain-builder/index.js
davecarter: to avoid linting warnings

kikoruiz: I know, but why do we avoid the linting warning and do not take another approach in the code below?

packages/sui-studio-utils/src/domain-builder/index.js
```diff
@@ -61,10 +62,18 @@ export default class DomainBuilder {
...
-                        ? successResponse(params)
+                        ? inlineError
+                          ? [null, successResponse(params)]
+                          : successResponse(params)
+                        : inlineError
+                        ? [null, successResponse]
                         : successResponse
                     )
-                  : Promise.reject(self._useCases[useCase].fail)
+                  : Promise.resolve(
```
midudev: 👀 This is a breaking change for people not using inlineError.

carlosvillu: That suppose to be a ternary too.

But it is true in this way it is a  breaking change.

packages/sui-studio-utils/src/domain-builder/index.js
```diff
@@ -61,10 +62,18 @@ export default class DomainBuilder {
                 return self._useCases[useCase].success !== undefined
                   ? Promise.resolve(
                       typeof successResponse === 'function'
-                        ? successResponse(params)
+                        ? inlineError
```
ferransimon: I see here tons of conditions but not a single test, it would be nice to have a test to, at least, check if it keeps the backward compatibility with the clients using the old version of the API

midudev:
In case you want to rethink the nested ternaries, you could do something like this:

```
const {fail, success} = self._useCases[useCase]
const { resolve, reject } = Promise

const data = typeof success === 'function'
  ? success(params)
  : success

const responseParams = data !== undefined
  ? { err: null, data }
  : { err: fail, data: null }

const createResponse = ({err, data}) => {
  if (inlineError)
    return resolve([err, data])

  return err
    ? reject(err)
    : resolve(data)
}

return createResponse(responseParams)
```

davecarter:
> In case you want to rethink the nested ternaries, you could do something like this (not tested :P):
> 
> ```
> const {fail, success} = self._useCases[useCase]
> const { resolve, reject } = Promise
> 
> const data = typeof success === 'function'
>   ? success(params)
>   : success
> 
> const responseParams = data !== undefined
>   ? { err: null, data }
>   : { err: fail, data: null }
> 
> const createResponse = ({err, data}) => {
>   if (inlineError)
>     return resolve([err, data])
> 
>   return err
>     ? reject(err)
>     : resolve(data)
> }
> 
> return createResponse(responseParams)
> ```

awwwwesome 😍😍😍 👍

PR #699, opened 2019-11-13
feat(sui-lint): upgrade dependencies and add prettier-stylelint

midudev:
Add prettier-stylelint and upgrade dependencies.


midudev:
Yes @rmoralp. I will create another PR. I want still to test this is working as expected in different computers and setups. And, as this is not a breaking change, I mantained the README for now until checked.

PR #700, opened 2019-11-14
feat(sui-js): add isTablet stat

jcger:
adds user-agent stats test and isTablet option


packages/sui-js/test/server/userAgentParserSpec.js
```diff
@@ -0,0 +1,2 @@
+import test from '../userAgentParserTest'
+test()
```
jcger: this is the only way I found to run the tests as server tests, alternatives are welcome

PR #702, opened 2019-11-18
Feat/upgrade lint and support config on package

midudev:
Nov-18-2019 15-46-56

- [x] **Fix stylelint and prettier integration on vscode.** 🔥 
- [x] New feature that adds automatically to the package.json the linter config needed. It will update the package.json **only if needed**. It will also remove deprecated sasslintConfig if present. 🤖 
- [x] Explain every prettier rule used. 👩‍🏫  
- [x] Improve README for configuring linter. ✍️  
- [x] Upgrade dependencies. 🆙 

**Important** 👀: Some package.json changes are needed as stated on the REAMDE. However, this version adds a new way to inject automatically the needed config to the package.json in order to make sure you have the needed config in the file. So, not breaking changes.

**Important 2**: Definitely you will have to RECHECK YOUR VSCODE CONFIG. Check the new README in order to put the proven correct config.


packages/sui-lint/.prettierrc.js
```diff
@@ -1,13 +1,22 @@
...
+  /* Specify the line length that the printer will wrap on. */
+  printWidth: 80,
+  /* Print semicolons at the ends of statements. */
+  semi: false,
+  /* Use single quotes instead of double quotes. */
+  singleQuote: true,
+  /* Specify the number of spaces per indentation-level. */
+  tabWidth: 2,
+  /* Print trailing commas wherever possible when multi-line.  */
+  trailingComma: 'es5',
```
jcger: 👀

midudev: Well, the code will work no matter this. In fact, the linter shouldn't give you an error with or without the trailing comma. But every-time prettier format your code, it will be formatted different (with trailing comma). It's a styling change and everything should keep work as expected.

midudev: I could remove the change to keep the old behavior. :) It's true some linter error could show up but prettier should fix automatically.

PR #703, opened 2019-11-19
feat(sui-js): add getRandomString function

rmoralp:
## Description
Add getRandomString in sui-js in order to have the possibility to create random strings with a defined length

## Example
https://github.com/SUI-Components/sui/pull/703/files#diff-e231c378ec264fa0c1d2439f65a9d2afR109

```
import {getRandomString} from '@s-ui/js/lib/string'

const randomStringLength = 6
const randomString = getRandomString(randomStringLength)

console.log(randomString.length) // log = 6 || 15 by default
console.log(randomString) // qwerty
```

### DIFF
--- a/packages/sui-js/README.md
+++ b/packages/sui-js/README.md
@@ -104,6 +104,14 @@ console.log(fromArrayToCommaQueryString({userId: 1, adId: 2, products: [3, 4, 5]
 import {htmlStringToReactElement} from '@s-ui/js/lib/string'
 
 htmlStringToReactElement('<p>No more dangerouslySetInnerHTML</p>')
+
+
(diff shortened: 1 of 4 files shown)


jcger:
just curious, what use are you going to give this?

jcger:
just curious, what use are you going to give this?

rmoralp:
We are loading one react component twice in same page as a widget in Monolito. We need that both of them have differents id's. When we continue with that page migration to react stack, there will be only one button and we won't use this function anymore for that. 
We know that it's better to have an uuid generator for an unique id creation but taking into account that it has a short life cycle we want to have a smart piece of code to do that.
You can see lead/callMeBack component in Motor repository.

PR #712, opened 2019-12-12
[Feature] Add @s-ui/react-router package

carlosvillu:
In order to move forward and stop using the deprecated _ContextTypes_, we're extracting _React Router 3_ core functionality with some tweaks so we don't need to push too many changes in our products but we could stop using deprecated functionalities.

Ongoing work 👷 :
- [x] Create @s-ui/react-router based on React Router 3.
- [x] Add README with client and server setup.
- [x] Extract React Router v3 API Documentation and adapt to our needs.
- [x] Add missing injected props
- - [x] location
- - [x] routeParams
- - [x] route
- [x] Test `<Redirect>` Component.
- [x] Add missing `<Redirect>` functionality on redirecting dynamic paths.
- [x] Test `<Link>` Component.
- [x] Fix `<Link>` rendering empty className when not needed.
- [x] Fix `<Link>` active classnames for pathes.
- [x] Test `<Router>` Component.
- [x] Fix `<Router>` creating new object on every render, causing perf issues on consumers.
- [x] Test `<Route>` Component.
- [x] Fix when no `<Route>` matches, it explodes.
- [x] Test `<IndexRoute>` Component.
- [x] Test utils (match, isActive and some more).
- [x] Adapt @s-ui/ssr to use new @s-ui/react-router.
- [x] Separate internal libs to a specific folder
- [x] Add missing support for using library on pure CSR.
- [x] Add hooks ↩️ 
- - [x] Document new hooks in README
- - [x] Add useLocation
- - [x] Add useParams
- - [x] Add useRouter
- - [x] Add Tests for new hooks
- - [x] Add aliases for @s-ui/bundler to fix linking.

**Features that won't be added for now and are missing from React Router v3:**
- [ ] Add missing `<Redirect>` functionality on composed paths.
- [ ] Add missing `<IndexRoute>` nested deeply support.
- [ ] Add missing `<IndexRedirect>` (used in sui-studio).


packages/sui-react-router/src/ReactRouterContext.js
```diff
@@ -1,17 +1,3 @@
-import React from 'react'
-import hoistNonReactStatics from 'hoist-non-react-statics'
+import {createContext} from 'react'
 
-const ReactRouterContext = React.createContext({})
-
-ReactRouterContext.wrapper = (Component, displayName) => {
```
carlosvillu: I am using it here: https://github.mpi-internal.com/scmspain/frontend-ma--uilib-components/pull/804/files#diff-4b99e5c03caaee681ecb10f20dcbaab4R173

midudev: Isn't a possibility to use `contextType` (the one that's not deprecated) to simplify it the library and the change you need to do until you the component is moved to a function?

```
AuthenticateComponent.contextType = RRContext
```

packages/sui-react-router/src/PatternUtils.js
```diff
@@ -0,0 +1,237 @@
...
+    regexpSource += '$'
+  }
+
+  const match = pathname.match(new RegExp(`^${regexpSource}`, 'i'))
+  if (match == null) {
+    return null
+  }
+
+  const matchedPath = match[0]
+  let remainingPathname = pathname.substr(matchedPath.length)
```
jcger, line 111:
```
Warning: Although String.prototype.substr(…) is not strictly deprecated (as in "removed from the Web standards"),
it is considered a legacy function and should be avoided when possible. It is not part of the core JavaScript
language and may be removed in the future. If at all possible, use the substring() method instead.
```

from the [docu](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/substr)

anyhow, isn't this the function with the memory leak?

carlosvillu: Good one!

carlosvillu:
Yes, @devictoribero this is just what we need to get the desired RegExp on the routes.

When we get it in production in all the sites and see that it is stable. The next step will be to consider how we would like to evolve the library.

Before we couldn't but now we have the control of the router internals. We could do much more than that. For example, optimize the load of the components and some of the styles.

packages/sui-react-router/package.json
```diff
@@ -0,0 +1,34 @@
+{
+  "name": "@s-ui/react-router",
+  "version": "1.0.0-beta.13",
```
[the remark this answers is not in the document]
midudev: As soon as it get released it should fix the version. :)

midudev:
@devictoribero before adding Regex I think we should check why React Router, the standard defacto is not using it and why we need them. To know better the specific case.

Tbh, I would avoid to walk away too much from different solutions as we could want to move back in order to avoid having to mantain such a big library. React Router v6 looks awesome, it's only 5KB (vs 15KB) and come with more features and, what's better, we won't mantain it but an entire community. :)

packages/sui-react-router/src/PatternUtils.js
```diff
@@ -0,0 +1,237 @@
...
+  return params
+}
+
+/**
+ * Returns a version of the given pattern with params interpolated. Throws
+ * if there is a dynamic segment of the pattern for which there is no param.
+ */
+export function formatPattern(pattern, params = {}) {
+  const {tokens} = compilePattern(pattern)
+  let parenCount = 0
```
jordevo, line 158: 👀 shouldn't it read `parentCount`?

packages/sui-react-router/src/PatternUtils.js
```diff
@@ -0,0 +1,237 @@
...
+  let pathname = ''
+  let splatIndex = 0
+  const parenHistory = []
```
jordevo, line 161: 👀 shouldn't it read `parentHistory`?

midudev: This whole file is a copy & paste 🍝 of react-router v3 as noted in the first line and thus is getting all the names for the variables.

I noticed as well but keep it in order to mantain the same code in case we need to recheck something from the original codebase.

packages/sui-react-router/src/PatternUtils.js
```diff
@@ -0,0 +1,237 @@
...
+        `Missing "${paramName}" parameter for path "${pattern}"`
+      )
+
+      if (paramValue == null) {
+        if (parenCount) {
+          parenHistory[parenCount - 1] = ''
+
+          const curTokenIdx = tokens.indexOf(token)
+          const tokensSubset = tokens.slice(curTokenIdx, tokens.length)
+          let nextParenIdx = -1
```
jordevo, line 206: 👀 `nextParentIdx` ?

packages/sui-react-router/src/PatternUtils.js
jordevo: oh got it! 👍

packages/sui-studio/package.json
```diff
@@ -33,7 +33,7 @@
     "bundle-loader": "0.5.6",
     "chai": "4.2.0",
     "chai-dom": "1.8.1",
-    "classnames": "2.2.6",
+    "classnames": "2.2.5",
```
nucliweb, line 36: We want to downgrade the `classnames` package?

midudev: Yes, to be sure we're using the same across. In other PR we could upgrade the classnames version in all the packages to be sure we load it only once.

PR #716, opened 2019-12-16
Simplify SUI Mono Phoenix

midudev:
- 🌴 **New** `no-root` option for phoenix, to do phoenix only for packages and not at root level.
- ☠️ **New** `no-audit`option for phoenix to avoid auditing dependencies when CI environment.
- 🏃 Simplify installation of packages and remove very CPU consuming listr dependency.
- 🕴 Less fancy installation, less noise from your computer fans.
- 🔀 Reliable concurrency to avoid problems with latest npm versions.
- 📑 Fix JSDoc wrong comments.
- 🏋 Package size: From 7MB to 13.7 kB by removing asset gif (faster installation!)

Untitled


carlosvillu:
Do you have an idea to do this, without the `cd demo` component?!
https://github.mpi-internal.com/scmspain/frontend-ma--uilib-components/blob/master/Makefile#L40-L41

midudev:
@carlosvillu further development would be needed in order to avoid the `cd demo`. We could think about adding a new param to provide the path where to start the commands.

packages/sui-mono/bin/sui-mono-phoenix.js
```diff
@@ -32,76 +38,49 @@ program
...
-              ? task.title + '\n' + '... work in progress ...'.yellow
-              : '... completed successfully!'.green + '\n'
+    queue
+      .add(() => serialSpawn(commands))
+      .then(() => {
+        if (progress) {
+          const {size, pending} = queue
+          const packageName = basename(cwd).grey
+          logUpdate(
+            `Installed ${packageName}. ${size + pending}/${scopes.length}`
```
adriadoz: IMHO, should be "All packages installed"

midudev: I agree the message could be improved.

With your message would be:
```
button. 23/45 installed
```

Now is:
```Installed button. 23/45```

What  do you think to something like this to make it more clear?
```button installed. 23 of 45 packages installations```

Comment (adriadoz) [reply packages/sui-mono/bin/sui-mono-phoenix.js]: `Button installed. Progress: 23 of 45 packages
button: 23 of 45 packages installed`

File: packages/sui-mono/package.json [hunk 2]
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@s-ui/mono",
-  "version": "1.44.0",
+  "version": "1.44.1-beta.3",
```
jcger: beta?

midudev: I'lll removed it before merging. It's in case I change something that needs to be tested :D

packages/sui-mono/package.json
```diff
@@ -21,7 +21,8 @@
     "commander": "4.0.1",
     "commitizen": "4.0.3",
     "conventional-changelog": "3.1.18",
-    "listr": "0.14.3",
+    "log-update": "3.3.0",
+    "p-queue": "6.2.1",
```
desko27: Good ones!

packages/sui-mono/bin/sui-mono-phoenix.js
midudev: Changed! Thanks @adriadoz

PR #718, opened 2019-12-19
fix(sui-ssr): avoid using the same name for different variables

kikoruiz:


packages/sui-ssr/server/criticalCss/index.js
```diff
@@ -68,7 +64,7 @@ export default criticalCSS => (req, res, next) => {
         logMessage(`Generation Critical CSS for -> ${urlRequest} with ${hash}`)
 
         const serviceRequestURL = `https://critical-css-service.now.sh/${device}/${urlRequest}`
-        const headers = criticalCSS.customHeaders
```
kikoruiz: This variable was also initialised in the line https://github.com/SUI-Components/sui/pull/718/files#diff-c9091fba6326a227b587bbfb9360ac12R61

PR #730, opened 2020-01-20
feat(sui-mockmock): add patch method to sui-mockmock

quinwacca:
## Description
Just add `PATCH` as a new method to mock in our tests.

## Example
```
const mocker = new Mocker()

mocker
        .httpMock(_API_BASE_URL)
        .patch(
          `${_API_UPDATE_VEHICLE_URL}?vehicleId=${VEHICLE_ID}`
        )
        .reply(updateUserVehicleUseCaseApiResponse, 200)
```


jcger:
what about adding a test?

PR #731, opened 2020-01-21
[sui-test]: Set properly the timeout in browser and server

carlosvillu:
This change is [Reviewable](https://reviewable.io/reviews/sui-components/sui/731)

### DIFF
--- a/packages/sui-test/bin/karma/index.js
+++ b/packages/sui-test/bin/karma/index.js
@@ -25,7 +25,14 @@ const createServer = config => {
   return server
 }
 
-module.exports = async ({watch, ci, pattern, ignorePattern, srcPattern}) => {
+module.exports = async ({
+  watch,
+  ci,
+  pattern,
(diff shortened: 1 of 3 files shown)


packages/sui-test/bin/karma/index.js
```diff
@@ -38,6 +45,11 @@ module.exports = async ({watch, ci, pattern, ignorePattern, srcPattern}) => {
     [pattern]: ['webpack']
   }
 
+  configure.client.mocha = {
+    ...configure.client.mocha,
+    ...(timeout && {timeout})
```
tomasmax: 👀 it would be enough passing just`timeout`

davecarter: ... or set a default value

carlosvillu: I am not sure what could happen if I pass timeout as undefined to the config.

PR #738, opened 2020-01-28
feat(sui-test): update cypress dep

AgonisticKatai:
## Description
Update `cypress` deps to prevent unexpected gygia error

## Related Issue
The current installed version of cypress, use the version **59** of `Electron`, that generates this unexpected behavior with Gygia's login. With the latest version of cypress, `Electron` **78** is installed, which solves this.

image


jordevo:
so long as it doesn't break anything... 🤞

kikoruiz:
@AgonisticKatai and what about the other Cypress versions defined in this package? I think they should be upgraded as well:
https://github.com/SUI-Components/sui/blob/f7faa2802541c641d1f7f8b60ff902cdb38f0330/packages/sui-test/Dockerfile#L5
https://github.com/SUI-Components/sui/blob/b805ee92a5d6c89e53b9b96218ec95bac51196b8/packages/sui-test/package.json#L4
Don't you think @carlosvillu ?

AgonisticKatai:
yep, it seems to show up in more places. I'm uploading the changes...

PR #742, opened 2020-01-31
[sui-studio ] Remove experimental flag

carlosvillu:
Automatic creation of all component´s test folders in the post-install of the studio

---
This change is [Reviewable](https://reviewable.io/reviews/sui-components/sui/742)


packages/sui-studio/scripts/postinstall.js
```diff
@@ -0,0 +1,84 @@
...
+  )
+
+const showError = msg => {
+  console.error(msg)
+  process.exit(1)
+}
+
+const writeFile = (path, body) => {
+  return new Promise((resolve, reject) => {
+    fse.outputFile(path, body, err => {
```
midudev: fs-extra already returns a Promise so you don't need to wrap it in one.

packages/sui-studio/scripts/postinstall.js
```diff
@@ -0,0 +1,84 @@
...
+      } else {
+        console.log(`Modified ${path}`)
+        resolve()
+      }
+    })
+  })
+}
+
+const createDir = path => {
+  return new Promise((resolve, reject) => {
```
midudev: Same. You could use the Promise returned instead wrapping it.

packages/sui-studio/scripts/postinstall.js
carlosvillu: Copy and paste power!

packages/sui-studio/scripts/postinstall.js
```diff
@@ -0,0 +1,81 @@
...
+
+// import React from 'react'
+// import {render} from '@testing-library/react'
+
+import chai, {expect} from 'chai'
+import chaiDOM from 'chai-dom'
+
+chai.use(chaiDOM)
+
+describe('AtomButton', () => {
```
midudev: Can we at least create the component name with the category and name of the component so we don't create always AtomButton?

packages/sui-studio/scripts/postinstall.js
```diff
@@ -0,0 +1,81 @@
...
+      .replace(process.env.INIT_CWD + '/components/', '')
+      .replace('/src/index.js', '')
+  )
+
+const showError = msg => {
+  console.error(msg)
+  process.exit(1)
+}
+
+const writeFile = async (path, body) => {
```
midudev: Not needed `async` method.

packages/sui-studio/scripts/postinstall.js
```diff
@@ -0,0 +1,81 @@
+const fse = require('fs-extra')
+const fg = require('fast-glob')
+
+console.log(`Looking for test folder in ${process.env.INIT_CWD}`)
```
midudev: Instead using always a environment variable, let's create a const here and reuse everywhere in case we want to recheck this in the future.

`const CWD = process.env.INIT_CWD`

packages/sui-studio/scripts/postinstall.js
```diff
@@ -0,0 +1,81 @@
...
+    .then(() => {
+      console.log(`Created ${path}`)
+    })
+    .catch(() => {
+      showError(`Fail creating ${path}`)
+    })
+}
+
+const BODY_TEST = `/**
+ * Remember: YOUR COMPONENT IS DEFINE GLOBALLY
```
midudev: Typo `DEFINE` -> `DEFINED`

PR #757, opened 2020-02-13
fix(sui-studio): sui-studio test breaks if we does not use patcher-mo…

rmoralp:
Given this code:
```
import React from 'react'

import chai, {expect} from 'chai'
import chaiDOM from 'chai-dom'
import {render} from '@testing-library/react'

chai.use(chaiDOM)

describe('FormColor', () => {
  it('Render', () => {
    render(<FormColor />)
    expect(true).to.be.eql(false)
  })
})
```
Sui-studio does not inject well the context:
image

### DIFF
--- a/packages/sui-studio/src/components/test/index.js
+++ b/packages/sui-studio/src/components/test/index.js
@@ -34,7 +34,7 @@ const Test = ({open, importTest, importComponent, context}) => {
       !EnhanceComponent.displayName &&
         console.error('[sui-Test] Component without displayName') // eslint-disable-line
       window[cleanDisplayName(EnhanceComponent.displayName)] = props => (
-        <SUIContext.Provider value={nextContext.default}>
+        <SUIContext.Provider value={nextContext}>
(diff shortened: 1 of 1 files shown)


packages/sui-studio/src/components/test/index.js
```diff
@@ -34,7 +34,7 @@ const Test = ({open, importTest, importComponent, context}) => {
       !EnhanceComponent.displayName &&
         console.error('[sui-Test] Component without displayName') // eslint-disable-line
       window[cleanDisplayName(EnhanceComponent.displayName)] = props => (
-        <SUIContext.Provider value={nextContext.default}>
+        <SUIContext.Provider value={nextContext}>
```
rmoralp: L:11 `const interOP = obj => (obj.default ? obj.default : obj)`
L:26 `let nextContext = interOP(context)`
 https://github.com/SUI-Components/sui/pull/757/files#diff-f33065125e6ed73b30d287a303ca88b1L26

PR #764, opened 2020-02-21
feat(sui-ssr): priorize head tag load for improve SEO

javiauso:
## Description
Priorize head tag load for improve SEO

### DIFF
--- a/packages/sui-ssr/server/template/index.js
+++ b/packages/sui-ssr/server/template/index.js
@@ -18,11 +18,17 @@ export const getTplParts = req => {
 export class HtmlBuilder {}
 
 HtmlBuilder.buildHead = ({headTplPart, helmetHead = {}}) => {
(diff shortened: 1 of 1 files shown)


packages/sui-ssr/server/template/index.js
```diff
@@ -18,11 +18,17 @@ export const getTplParts = req => {
 export class HtmlBuilder {}
 
 HtmlBuilder.buildHead = ({headTplPart, helmetHead = {}}) => {
+  const headElement = headTplPart.substr(headTplPart.indexOf('<head'))
```
midudev: Is this because we expect somebody would put attributes in the `<head>` element?

javiauso: Yes, for example in Infojobs the head element have some attrributte

midudev: Thanks!

PR #765, opened 2020-02-25
docs(sui-studio): update docs with known issues

rmoralp:


packages/sui-studio/README.md
```diff
@@ -171,6 +171,31 @@ describe.context.other('atom/button', AtomButton => {
 })
 ```
 
+### Known issue: Test a memoized component
+
+If a component is exported wrapped memoized: `export default React.memo(Component)`, it loses the displayName and sui-test dispatch an Error because of couldn't find the component.
```
midudev: ` because of couldn't find the component.` -> `because it couldn't...`

PR #766, opened 2020-03-03
fix(sui-domain): make subscribe method take onNext, onError arguments

jordevo:
Fixes the subscribe behaviour on useCases using the `@streamify` decorator.

## Description
So far, only the first argument was taken by the subscribe method (which happened to be `onNext` from `@streamify`). The method is now also taking `onError` for a correct error handling. 

## Related Issue
N/A

## Example
N/A

### DIFF
--- a/packages/sui-domain/src/EntryPointFactory.js
+++ b/packages/sui-domain/src/EntryPointFactory.js
@@ -48,7 +48,7 @@ export default ({useCases, config}) =>
             },
             $: {
               execute: {
-                subscribe: fn => {
+                subscribe: (onNext,onError) => {
                   // creating an object here that will have a dispose method
                   const ret = {dispose: function() {}}
                   loader().then(factory => {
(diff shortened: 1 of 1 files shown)


packages/sui-domain/src/EntryPointFactory.js
```diff
@@ -48,15 +48,15 @@ export default ({useCases, config}) =>
...
+                subscribe: (onNext,onError) => {
                   // creating an object here that will have a dispose method
                   const ret = {dispose: function() {}}
                   loader().then(factory => {
                     // black magic: mutate the object, very small memory leak but that
                     // makes dispose working async and we need it
                     ret.dispose = factory.default[method]({
                       config: this._config
-                    }).$.execute.subscribe(fn).dispose
+                    }).$.execute.subscribe(onNext,OnError).dispose
```
AgonisticKatai: OnError => onError 🔎

PR #767, opened 2020-03-04
feat(sui-js): add encode to toQueryString options

jordevo:
…and subsequent call to `stringify`

## Description
One of our use cases uses toQueryString to pass on a URL which doesn't need to be encoded. Hence, we need to be able to pass the option `encode` as `false` to `qs`'s `stringify`. 

## Related Issue
N/A

## Example
N/A

### DIFF
--- a/packages/sui-js/src/string/to-query-string.js
+++ b/packages/sui-js/src/string/to-query-string.js
@@ -12,9 +12,9 @@ function toQueryString(queryParams, options = {}) {
   const {arrayFormat, delimiter, encode = true} = options
 
   const mergedOptions = {
+    encode,
     ...(typeof arrayFormat !== 'undefined' && {arrayFormat}),
-    ...(typeof delimiter !== 'undefined' && {delimiter}),
-    ...(typeof encode !== 'undefined' && {encode})
+    ...(typeof delimiter !== 'undefined' && {delimiter})
   }
 
   return stringify(queryParams, mergedOptions)
--- a/packages/sui-js/src/string/to-query-string.js
+++ b/packages/sui-js/src/string/to-query-string.js
@@ -9,7 +9,7 @@ import {stringify} from 'qs'
  * @param {string} [options.delimiter] - delimiter
  */
 function toQueryString(queryParams, options = {}) {
-  const {arrayFormat, delimiter, encode} = options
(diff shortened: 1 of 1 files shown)


AitorRodriguez990:
@giodelabarrera

packages/sui-js/src/string/to-query-string.js
```diff
@@ -9,11 +9,12 @@ import {stringify} from 'qs'
  * @param {string} [options.delimiter] - delimiter
```
giodelabarrera:
```suggestion
 * @param {string} [options.delimiter] - delimiter
 * @param {boolean} [options.encode=true] - encode
```

packages/sui-js/src/string/to-query-string.js
```diff
@@ -9,11 +9,12 @@ import {stringify} from 'qs'
...
  */
 function toQueryString(queryParams, options = {}) {
-  const {arrayFormat, delimiter} = options
+  const {arrayFormat, delimiter, encode} = options
```
giodelabarrera: I believe that in this case we must put a default value because we are creating our API. So we do not leave that default responsibility to the library on we use, in this case `qs`.

```suggestion
  const {arrayFormat, delimiter, encode = true} = options
```

packages/sui-js/src/string/to-query-string.js
```diff
@@ -9,11 +9,12 @@ import {stringify} from 'qs'
...
 
   const mergedOptions = {
     ...(typeof arrayFormat !== 'undefined' && {arrayFormat}),
-    ...(typeof delimiter !== 'undefined' && {delimiter})
+    ...(typeof delimiter !== 'undefined' && {delimiter}),
+    ...(typeof encode !== 'undefined' && {encode})
```
giodelabarrera: Change this line by this:

```js
 const mergedOptions = {
    encode,
    ...(typeof arrayFormat !== 'undefined' && {arrayFormat}),
    ...(typeof delimiter !== 'undefined' && {delimiter})
...
```

packages/sui-js/src/string/to-query-string.js
tureey: +1

PR #774, opened 2020-03-10
feat(sui-js): add highlightText method to sui-js

arnau-rius:
Add highlightText method to sui-js

## Description
highlightText method allows to add a `startTag` and `endTag` around all ocurrences of certain `query` of a `value` in a case-insensitive and accent-insensitive way.

## Related Issue
https://jira.scmspain.com/browse/SUIC-447

## Example
Tests and Documentation updated (see examples)
If value is `Cádiz`, query is `ca`, startTag is `<strong>`, endTag is `</strong>` it returns: `<strong>Cá</strong>diz`

### DIFF
--- a/packages/sui-js/src/string/highlighter.js
+++ b/packages/sui-js/src/string/highlighter.js
@@ -1,0 +1,47 @@
+import {remove as removeAccents} from 'remove-accents'
+
+// given a value and a query, if value has the query as a subset highlight it using startTag and endTag
+export const highlightText = ({value, query, startTag, endTag}) => {
+  // replace accents and lowercase
+  const queryWithoutAccents = removeAccents(query).toLowerCase()
+  const valueWithoutAccents = removeAccents(value).toLowerCase()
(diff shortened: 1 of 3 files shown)


packages/sui-js/src/string/highlighter.js
```diff
@@ -0,0 +1,47 @@
...
+export const highlightText = ({value, query, startTag, endTag}) => {
+  // replace accents and lowercase
+  const queryWithoutAccents = removeAccents(query).toLowerCase()
+  const valueWithoutAccents = removeAccents(value).toLowerCase()
+  // get all the matches of the query in the value
+  const matches = findMatches(valueWithoutAccents, queryWithoutAccents)
+
+  // highlight the matches of the query in the value
+  const highlighterReducer = (valueHighlightedAccumulator, match, i) => {
+    const start = match.start + i * (startTag + endTag).length
```
midudev: `(startTag + endTag).length` could be calculated once.

PR #777, opened 2020-03-13
fix(sui-mockmock): fix restore interceptors and use new nock version

rmoralp:
## Description
In all verticals we are restoring our mocks after each test:
image

But sui-mockmock does not use nock implementation for clear all the interceptors created by other test mocks. This PR add this implementation for clean all the interceptors when we use `mocker.restore()`

See docs 👀 https://github.com/nock/nock#restoring

## Issue
In motor we had a test that mocks `/authentication` with `reply({})`.
After this one, we have another test with same mock but with `reply({...someData})`
Given we never restore the interceptors, second test fails because mock replies with `{}` because of nock uses `interceptors.find` it appears earlier in interceptors array.

Just 🤯

## Summary
image

### DIFF
--- a/packages/sui-mockmock/src/http/serverMocker.js
+++ b/packages/sui-mockmock/src/http/serverMocker.js
@@ -2,6 +2,10 @@ import {Mocker, Mock} from './mockerInterface'
 import nock from 'nock'
 
 class ServerMocker extends Mocker {
+  restore() {
+    nock.cleanAll()
+  }
+
   httpMock(baseUrl) {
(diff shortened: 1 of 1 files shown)


rmoralp:
Our `mocker.restore` was calling our """"interface""""

image

image

We have same problem with `mocker.create()` but IMHO it has no sense to keep anymore. What do you think? @midudev @carlosvillu @jcger @kikoruiz

jordevo:
legendary 🙌

PR #783, opened 2020-03-26
feat(sui-widget-embedder): add new param hrefRegExp to identify the h…

izeller:
## Description
We need to download a widget for this kind of  url: _https://adevinta-spain.ofertas-trabajo.infojobs.net/_
The allowed regex to download widgets are based on pathname but it does not allow to get the whole href.
The idea is to add a new param to download the widget with a regex based on href.

### DIFF
--- a/packages/sui-widget-embedder/downloader/index.js
+++ b/packages/sui-widget-embedder/downloader/index.js
@@ -108,12 +108,12 @@
   }
 
   function matchNameWithRegExp(name, regExp) {
-    if(!regExp) {
+    if (!regExp) {
       return false
     }
     if (Array.isArray(regExp)) {
       return regExp.some(function(re) {
-        name.match(new RegExp(re))
+        return name.match(new RegExp(re))
       })
     }
 
--- a/packages/sui-widget-embedder/downloader/index.js
+++ b/packages/sui-widget-embedder/downloader/index.js
@@ -107,24 +107,38 @@
     })
   }
 
-  function matchPathnameWithRegExp(regExp) {
-    var pathname = window.location.pathname
+  function matchNameWithRegExp(name, regExp) {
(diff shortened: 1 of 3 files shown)


packages/sui-widget-embedder/downloader/index.js
```diff
@@ -107,24 +107,38 @@
...
-  function matchPathnameWithRegExp(regExp) {
-    var pathname = window.location.pathname
+  function matchNameWithRegExp(name, regExp) {
+    if(!regExp) {
+      return false
+    }
     if (Array.isArray(regExp)) {
       return regExp.some(function(re) {
-        pathname.match(new RegExp(re))
+        name.match(new RegExp(re))
```
jcger: i don't know how this was working before, you must return this value

midudev: Good catch. It should be something like this:
`return regExp.some(re => name.match(new RegExp(re)))`

packages/sui-widget-embedder/downloader/index.js
```diff
@@ -107,24 +107,38 @@
     })
   }
 
-  function matchPathnameWithRegExp(regExp) {
-    var pathname = window.location.pathname
+  function matchNameWithRegExp(name, regExp) {
+    if(!regExp) {
```
midudev: I think the linter should warn you about this line to be `if (!regExp) {` instead `if(!regExp) {`

izeller: Fixed

packages/sui-widget-embedder/downloader/index.js
izeller: Changed! I don't know but it was working

PR #793, opened 2020-04-06
feat(sui-test): add ci mode and update cypress version

jcger:
BREAKING CHANGE:
update cypress version

### DIFF
--- a/packages/sui-test/Dockerfile
+++ b/packages/sui-test/Dockerfile
@@ -2,6 +2,8 @@ FROM cypress/base
 
 WORKDIR /usr/src
 
-RUN npm install --force-only @s-ui/test
+ARG CYPRESS_VERSION=4.3.0
+
+RUN npm install --force-only @s-ui/test cypress@$CYPRESS_VERSION
 
 ENTRYPOINT ["npx", "sui-test", "e2e"]
--- a/packages/sui-test/Dockerfile
+++ b/packages/sui-test/Dockerfile
@@ -2,8 +2,6 @@ FROM cypress/base
 
 WORKDIR /usr/src
 
-ARG CYPRESS_VERSION=3.8.3
-
-RUN npm install --force-only @s-ui/test cypress@$CYPRESS_VERSION
+RUN npm install --force-only @s-ui/test
 
 ENTRYPOINT ["npx", "sui-test", "e2e"]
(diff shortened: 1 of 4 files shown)


packages/sui-test/bin/sui-test-e2e.js
```diff
@@ -86,6 +88,11 @@ if (screenshotsOnError) {
   cypressConfig.screenshotsFolder = SCREENSHOTS_FOLDER
 }
 
+if (ci) {
```
midudev, line 91: 😮 mola.

packages/sui-test/Dockerfile
```diff
@@ -2,8 +2,6 @@ FROM cypress/base
 
 WORKDIR /usr/src
 
-ARG CYPRESS_VERSION=3.8.3
-
-RUN npm install --force-only @s-ui/test cypress@$CYPRESS_VERSION
+RUN npm install --force-only @s-ui/test
```
carlosvillu, line 5: Why are you remove the installation of Cypress ?!

jcger: didn't catch at first that this is to avoid lazy load the package, isn't it? gonna roll it back

tomcask: If we remove cypress, where will it be then? In the base image of docker there is no cypress.

In the [documentation](https://hub.docker.com/r/cypress/base/) it recommends using this:

`FROM cypress/base:10
RUN npm install --save-dev cypress
RUN $(npm bin)/cypress verify
RUN $(npm bin)/cypress run`

packages/sui-test/README.md
```diff
@@ -135,6 +135,7 @@ Then you can use in your specs `cy.login()`
     -U, --userAgentAppend <userAgentAppend>  Append string to UserAgent header.
     -UA, --userAgent <userAgent>             Overwrite string to UserAgent header.
     -G, --gui                                Run the tests in GUI mode.
+    -C, --ci                                 CI Mode, reduces memory consumption
```
nucliweb, line 138: 👏🏼

packages/sui-test/Dockerfile
jcger: it's being lazy loaded by the sui-test itself https://github.com/SUI-Components/sui/blob/master/packages/sui-test/bin/sui-test-e2e.js#L89

tomcask: ok!!!

tomcask: 🎉

PR #796, opened 2020-04-17
[sui-ssr] Use parse to improve JSON

carlosvillu:
This change is [Reviewable](https://reviewable.io/reviews/sui-components/sui/796)

### DIFF
--- a/packages/sui-ssr/server/template/index.js
+++ b/packages/sui-ssr/server/template/index.js
@@ -67,8 +67,20 @@ HtmlBuilder.buildBody = ({
   return html
 }
 
-HtmlBuilder.injectDataHydration = ({windowPropertyName, data = {}}) =>
-  `<script>window.${windowPropertyName} = ${JSON.stringify(data).replace(
-    /<\//g,
-    '<\\/'
-  )};</script>`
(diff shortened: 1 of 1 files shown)


packages/sui-ssr/server/template/index.js
```diff
@@ -67,8 +67,20 @@ HtmlBuilder.buildBody = ({
...
+    return undefined // Invalid JSON
+  }
+
+  return stringified.replace(/\u2028|\u2029/g, str =>
+    str === '\u2029' ? '\\u2029' : '\\u2028'
+  ) // invalid in JavaScript but valid JSON
+}
+
+HtmlBuilder.injectDataHydration = ({windowPropertyName, data = {}}) => {
+  const jsonSource = JSON.stringify(stringifySafe(data))
```
midudev, line 83: This is transforming the JSON to string twice.

carlosvillu: Looks like that this is the intention originally: https://github.com/gfx/webpack/blob/1cc9f8799bd60daa0b01518294de8974a0fed495/lib/JsonGenerator.js#L50

1.- Generate a JSON string save
2.- Generate a JSON string from the JSON string save.

Looks like a double-check of the integrity of the string

PR #800, opened 2020-04-23
[sui-ssr] send hints preload with headers

carlosvillu:
Using Header, we can avoid waiting for the getInitialProps resolution to start fetching the critical assets.

---
This change is [Reviewable](https://reviewable.io/reviews/sui-components/sui/800)

### DIFF
--- a/packages/sui-ssr/server/prpl/index.js
+++ b/packages/sui-ssr/server/prpl/index.js
@@ -23,7 +23,7 @@ export default config => (req, res, next) => {
     __CACHE__ = {}
   }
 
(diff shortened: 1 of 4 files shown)


packages/sui-ssr/server/prpl/index.js
```diff
@@ -97,6 +97,7 @@ export default config => (req, res, next) => {
             __REQUESTING__ = false
             try {
               __CACHE__[hash] = JSON.parse(json)
+              console.log({json})
```
midudev, line 100: 👀

carlosvillu: We want to log this to kibana

PR #808, opened 2020-04-29
feat(sui-studio): add mocha global to test url redirects

arnau-rius:
## Description
add a mocha variable `__ROUTER_PUSH_TRIGGERED__ `, so component tests don't fail due to [`global leak detected error`](https://github.com/mochajs/mocha/issues/110).

## Example:

In context.js file we can mock the `router.push` function with: 
image

then in the test asserting that the url is in the window variable:

image

### DIFF
--- a/packages/sui-studio/src/index.html
+++ b/packages/sui-studio/src/index.html
@@ -8,7 +8,7 @@
   <script class="mocha-init">
     mocha.setup('bdd');
     mocha.checkLeaks();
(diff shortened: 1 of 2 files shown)


arnau-rius:
> Could you document this in a new section of the studio readme ?!
> 
> Maybe we can create a tip&trick section on testing component.

Good idea! 👍🏼📄

PR #813, opened 2020-05-04
Feat/move studio to new sui react router

midudev:
Before rolling the new react-router, let's eat our own dog food by using the new react-router in our studio.

- [x] Adapt routes to use `<Redirect>` instead `<IndexRedirect />`.

**Note:**
Thanks to testing the router in SUI-Studio we have detected some missing functionality regarding using the Router without `match` and relying only on client-side:
https://github.com/SUI-Components/sui/pull/712/commits/2b86b28b047be277ce7ab68c507bfc2d10498901

### DIFF
--- a/packages/sui-studio/package.json
+++ b/packages/sui-studio/package.json
@@ -27,7 +27,7 @@
     "@s-ui/helpers": "1",
     "@s-ui/mono": "1",
     "@s-ui/react-context": "1",
-    "@s-ui/react-router": "^1.0.0-beta.9",
+    "@s-ui/react-router": "^1.0.0-beta.10",
     "@testing-library/react": "10.0.4",
     "@testing-library/user-event": "10.0.1",
     "bundle-loader": "0.5.6",
--- a/packages/sui-studio/package.json
+++ b/packages/sui-studio/package.json
@@ -46,7 +46,7 @@
     "just-kebab-case": "1.1.0",
     "just-pascal-case": "1.1.0",
     "normalize.css": "8.0.1",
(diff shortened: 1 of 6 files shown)


packages/sui-studio/src/components/root/index.js
```diff
@@ -1,6 +1,6 @@
 import React from 'react'
-import {Router, browserHistory} from 'react-router'
+import {Router} from '@s-ui/react-router'
 import routes from '../../routes'
 import '../../index.scss'
 
-export default () => <Router routes={routes} history={browserHistory} />
+export default () => <Router routes={routes} />
```
midudev, line 6: `history` now is optional. It detects by default if we're on client or server in order to use the correct `history` for us.

packages/sui-studio/src/routes.js
```diff
@@ -12,9 +12,12 @@ import TestPage from './components/test-page/index'
 
 export default (
   <Route>
+    <Redirect
```
midudev, line 15: Sadly, we've removed support for nested Redirects and `IndexRedirect`. This is the way to keep the same functionality.

packages/sui-studio/package.json
```diff
@@ -27,6 +27,7 @@
     "@s-ui/helpers": "1",
     "@s-ui/mono": "1",
     "@s-ui/react-context": "1",
+    "@s-ui/react-router": "^1.0.0-beta.10",
```
midudev, line 30: Using the beta with caret to get new betas if available. We will change to major as soon as it's stable.

PR #817, opened 2020-05-06
feat(sui-lint): add warnings for a11y linter rules

midudev:
## Description
Add a11y recommended rules as warnings.

## Related Issue
In order to step to the right direction but keeping our builds to work as expected.

## Example
image


jordevo:
👏

AgonisticKatai:
thx! 🔥

PR #822, opened 2020-05-07
fix(sui-studio): stop having super large windows when errors are long

devictoribero:
When having an error, if the error is large, it disappears from the window and it's quite annoying because most of the time you can't read the message.

To solve this, you have to force a width in the `dev tools` of that element, and these styles disappear after applying a change in the test.

## Description
Give a `max-width` to the errors and test window to avoid this.

### Before
Screen Shot 2020-05-07 at 15 19 27

### After
Screen Shot 2020-05-07 at 15 20 22


jordevo:
👏

PR #829, opened 2020-05-14
feat(sui-mockmock): get nth request (url, body, headers) from clientMocker

arnau-rius:
- get nth request (url, body, headers) from clientMocker

## Description
Inside your client tests, now you can get a request mocked by sui-mockmock and inspect it.
For example you want to assert that the body of a certain request has a concrete body in order to test that your usecase is doing the transformations in a properly way

## Example
- First rectangle gets the first request : `const request = mock.requestNTH(0)`
-Second  rectangle asserts that the  request body is equal to a super long string allocated in a variable. `expect(request.body).to.be.equal(category245Ad.body)`

image

## Remain work
Implement both `toStandardRequest` and `requestNTH` in the serverMocker.js.

### DIFF
--- a/packages/sui-mockmock/README.md
+++ b/packages/sui-mockmock/README.md
@@ -1,4 +1,5 @@
 # sui-mockmock
+
 > Mocking utilities for testing.
 
 ## Motivation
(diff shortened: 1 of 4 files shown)


arnau-rius:
> Cool! Could you add some documentation to readme file?
> Why it is just only for client side? Does it have sense to be also supported for server side?

Yes, as the Remain work part of the description says, this PR adds support to client requests mocked. If this methods are used in serverside the interface `mockerInterface.js` will warn you that are not implemented already. So yes, it makes total sense (feel free 😜) !

Good point, adding documentation!! 👍

PR #832, opened 2020-05-21
feat(sui-studio): add scroll to the tests section

devictoribero:
When having some tests and/or errors that overflow the window, we have no scroll.

## Implementation
* Made max-width of the test window wider
* Added scroll in case of overflow

## Example 
Screen Shot 2020-05-21 at 09 26 06

### DIFF
--- a/packages/sui-studio/src/components/test/index.scss
+++ b/packages/sui-studio/src/components/test/index.scss
@@ -7,7 +7,8 @@
   &--open,
   &--failures {
     display: block;
-    max-width: 500px;
+    max-width: 700px;
+    overflow-y: auto;
   }
 }
 


packages/sui-studio/src/components/test/index.scss
```diff
@@ -7,7 +7,8 @@
   &--open,
   &--failures {
     display: block;
-    max-width: 500px;
+    max-width: 700px;
```
midudev, line 10: Fine tunning. XD

devictoribero: 😆 Time ago when I added this I would that it was okay, but the truth is, that for long tests it's not wider enough

davecarter: I **never** do so many errors to reach this height.. but if you do.. It's ok :trollface:

devictoribero: well, but I guess you test a lot, don't you? :trollface:

davecarter: Screenshot 2020-05-21 at 10 40 36

PR #833, opened 2020-05-21
Update deprecated dependencies in: sui-component-dependencies

xavicomi:
Update deprecated dependencies in: sui-component-dependencies

## Description
Add @s-ui/i18n
Delete @schibstedspain dependencies


midudev:
Before merging, let's make sure this is not going to be problematic with any vertical.

midudev:
Vamos a seguir otra estrategia para eliminar definitivamente este paquete. :)

PR #835, opened 2020-05-26
Fix/cache decorator error response

tomasmax:
## Description

- Improved cache key to be target constructor name + fnName.

- Don't apply cache for error responses when combining with @inlineError decorator.

### DIFF
--- a/packages/sui-decorators/src/decorators/cache/index.js
+++ b/packages/sui-decorators/src/decorators/cache/index.js
@@ -20,13 +20,15 @@ const _cache = ({
   target,
   ttl
 } = {}) => {
-  caches[`${target.constructor.name}::${fnName}`] =
-    caches[`${target.constructor.name}::${fnName}`] ||
+  const cacheKey = `${target.constructor.name}::${fnName}`
+
+  caches[cacheKey] =
+    caches[cacheKey] ||
     (algorithm === ALGORITHMS.LRU
       ? new LRU({size})
       : new Error(`[sui-decorators::cache] unknown algorithm: ${algorithm}`))
 
-  const cache = caches[`${target.constructor.name}::${fnName}`]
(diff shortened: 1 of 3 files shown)


packages/sui-decorators/src/decorators/cache/index.js
```diff
@@ -20,13 +20,13 @@ const _cache = ({
   target,
   ttl
 } = {}) => {
-  caches[fnName] =
-    caches[fnName] ||
+  caches[`${target.constructor.name}::${fnName}`] =
```
midudev, line 23: As the cacheKey now is a bit more complex, we might want to create it once.

```
const cacheKey = `${target.constructor.name}::${fnName}`
const cacheEntry = caches[cacheKey] || 
  (algorithm === ALGORITHMS.LRU
    ? new LRU({size})
    : new Error(`[sui-decorators::cache] unknown algorithm: ${algorithm}`))

caches[cacheKey] = cacheEntry
```

davecarter: I guess is not relevant in this case to build a `cacheKey` that Babel will transpile the final name of `target.constructor.name`

packages/sui-decorators/test/server/cacheSpec.js
```diff
@@ -3,51 +3,136 @@
...
 
-describe('Cache', () => {
-  it('should ignore the cache in Node by default', () => {
-    class Buz {
-      constructor() {
-        this.rnd = () => Math.random()
+describe('Cache in the server', () => {
+  describe('decorating a sync method', () => {
+    it('should not apply cache in Node by default', () => {
+      class Buz {
```
davecarter, line 11: You mean `Buzz`, right? 😄

PR #855, opened 2020-06-16
feat(sui-bundler): add statics cache only option

jcger:
## Description
Adds an option to activate static file cache but without having to create the offline.html page


packages/sui-bundler/README.md
```diff
@@ -196,7 +196,7 @@ import {register, unregister} from '@s-ui/bundler/registerServiceWorker'
 register({
   first: () => window.alert('Content is cached for offline use.'),
   renovate: () => window.alert('New content is available; please refresh.')
-})();
+});
```
midudev, line 199: Bien visto! 👏

PR #857, opened 2020-06-16
feat(sui-widget-embedder): allow multiple containers for single Widget

midudev:
We've found in Fotocasa that sometimes we want to render the same Widget in, for example, a list of cards. So we're adding a way to render the same component over different HTML nodes while keeping the old compatibility.

- [x] Deprecate the old `node` prop was not correct semantically and now is misleading.
- [x] Allow multiple containers by using a new prop `selector`.
- [x] Add new prop `renderMultiple` to render the widget to all the nodes found using the `selector`. 
- [x] Update README with the new props and fixing old as well.

### DIFF
--- a/packages/sui-widget-embedder/src-react/Widget.js
+++ b/packages/sui-widget-embedder/src-react/Widget.js
@@ -54,9 +54,13 @@ export default function Widget({
 
 Widget.propTypes = {
   browser: PropTypes.object,
-  children: PropTypes.element,
+  children: PropTypes.element.isRequired,
   domain: PropTypes.object,
   i18n: PropTypes.object,
   isVisible: PropTypes.bool,
-  node: PropTypes.string
+  node: () =>
+    new Error(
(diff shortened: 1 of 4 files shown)


packages/sui-widget-embedder/src-react/Widget.js
```diff
@@ -1,56 +1,62 @@
...
-  render() {
-    return null
-  }
+Widget.propTypes = {
+  browser: PropTypes.object,
+  children: PropTypes.element,
+  domain: PropTypes.object,
+  i18n: PropTypes.object,
+  isVisible: PropTypes.bool,
+  node: PropTypes.string
```
nucliweb, line 61: As I see in the documentation, the props `children`, `isVisible` and `selector` _(aka node)_ are required

midudev: I've improved the PropTypes and even added a deprecation to the `node` prop. :D

PR #859, opened 2020-06-17
feat(sui-i18n): add percentage formatting capability to i18n

jordevo:
Currently on our `i18n` we don't provide with a method to format percentages. 

## Description
This `PR` adds a new method `.p()` that adds a percentage sign to the number we pass on as a value.

## Example
`console.log(i18n.p(69));`

would tipically output:
`69 %`

### DIFF
--- a/packages/sui-i18n/src/i18n.js
+++ b/packages/sui-i18n/src/i18n.js
@@ -107,6 +107,14 @@ export default class Rosetta {
     throw new Error(`Invalid type '${type}' passed to i18n.f`)
   }
 
+  formatPercentage(value) {
+    return this.f('percentage', value)
+  }
+
+  formatPhone(phoneNumber) {
+    return this.f('phone', phoneNumber)
+  }
+
   url(urlPattern, allowQueryParams) {
     return urlPattern
       .split('/')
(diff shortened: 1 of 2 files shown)


rmoralp:
Could you add some test? 
https://github.com/SUI-Components/sui/blob/master/packages/sui-i18n/test/i18nSpec.js

jordevo:
hi @desko27 ! 

yes I thought of doing that instead but since percentages are part numbers and hence they're also localisable, I thought it'd be useful to have a similar approach to that of `.c` and `.n` ...

I didn't see any trade-off on that approach, is there any in any way?

desko27:
@jordevo sorry I didn't explain properly.

The point I meant is the existing API: why `p` should be used for `percent` and not for `phone` which already exists in `f`?

Apart from that, simplicity of `p` can't be denied. Both solutions are ok for me, if no one has more thoughts to share 😄

midudev:
Tbh, the `p` seems to be, at least, misleading. We're not forced to use a single char for methods on i18n. I would recommend to use as well the `f` function.

Maybe we could start exporting then properly named methods for the rest, like:

i18n.formatPercentatge() -> internally calls the `f` function.

PR #864, opened 2020-06-19
[sui-studio] Works with sui-react-router

carlosvillu:
Improve the integration of the sui-studio with sui-react-router To be able to develop and test the components.

Now it is still possible to pass a FakeRouter from the context.js file and it will be used inside the component as the current Router instance ignoring the one generated on the page.

---
This change is [Reviewable](https://reviewable.io/reviews/sui-components/sui/864)


packages/sui-studio/workbench/src/app.js
```diff
@@ -64,15 +65,30 @@ const importAll = requireContext => requireContext.keys().map(requireContext)
 
   const {raw} = params
   const ComponentToRender = raw ? Raw : Root
+
+  const routes = (
+    <Router>
```
midudev, line 70: You don't need to wrap the `routes` with another <Router> that's in fact a Context Provider.

packages/sui-react-router/src/Router.js
```diff
@@ -82,7 +83,7 @@ const Router = ({
   }, []) // eslint-disable-line
 
   return (
-    <RouterContext.Provider value={state}>
+    <RouterContext.Provider value={{...state, ...(fakeRouter && {router: fakeRouter})}}>
```
midudev, line 86: This could be problematic on production code as we're creating a new object in every render of this provider. In fact, there's already problems with the line 77 because of that.

midudev: Can we use the `router` prop to essentially do the same that the `fakeRouter` is trying to do?

carlosvillu: I preferred not to change the router until just passing it to the Provider, because I am not very clear what consequences it could have if there is some kind of navigation in the studio.

To avoid what you are saying, the best thing would be to mutate the state just before assigning it to the provider. That way we wouldn't create a new object every time.

packages/sui-studio/workbench/src/app.js
carlosvillu: I know, I tried to follow the same pattern as the tests, but for some reason it didn't work.

Until I wrapped everything up in a second Router always seemed like a blank page.

PR #871, opened 2020-06-30
Sui mono/single scope

jcger:
## Description
Adds the `--scope` param in order to just phoenix one package

## Example
`npx @s-ui/mono phoenix --no-root --scope atom/button` will delete atom/button's node_modules folder and package-lock.json file and run `npm install` in it

### DIFF
--- a/packages/sui-mono/bin/sui-mono-phoenix.js
+++ b/packages/sui-mono/bin/sui-mono-phoenix.js
@@ -49,6 +49,10 @@ program
     '--no-progress',
     'Force to not show progress of tasks (perfect for CI environments)'
   )
+  .option(
+    '-s, --scope <string>',
+    'Runs phoenix on a given scope, for example: -s atom/button'
+  )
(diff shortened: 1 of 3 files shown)


packages/sui-mono/bin/sui-mono-phoenix.js
```diff
@@ -49,6 +49,10 @@ program
     '--no-progress',
     'Force to not show progress of tasks (perfect for CI environments)'
   )
+  .option(
+    '-s, --scope <string>',
```
carlosvillu, line 53: You could make this an array, so you could spend more than one scope here.

ex.: https://github.com/SUI-Components/sui/blob/master/packages/sui-bundler/bin/sui-bundler-dev.js#L38-L46

jcger: Lets remind this as a possible future iteration as it's on top of what this task was meant to acomplish

PR #876, opened 2020-07-01
Feature/debounce

seremejvaz:
Debounce

## Description
Add debounce function from lodash.


andresin87:
what about creating a `debounce.js` file inside function directory and importing it in `index`. Its better for bundeling purposes if you only want to use the debounce fn and not all the functions that index will include

packages/sui-js/src/function/index.js
```diff
@@ -0,0 +1,2 @@
+import {debounce} from './debounce'
```
andresin87, line 1: ES6 supports directly exporting imported modules

PR #877, opened 2020-07-02
fix(sui-ssr): fix node version to 14-alpine in dockerfile.tpl

igncalleja:
## Description
fix node version to 14-alpine in dockerfile.tpl

## Background
Deploy pipelines are blocked because the `pm2:latest-alpine` is installing the node v 14.5.0 and this version is making pods fail when they try to start.

### DIFF
--- a/packages/sui-ssr/archive/Dockerfile.tpl
+++ b/packages/sui-ssr/archive/Dockerfile.tpl
@@ -1,4 +1,4 @@
-FROM keymetrics/pm2:latest-alpine
+FROM keymetrics/pm2:14-alpine
 
 WORKDIR /usr/src/app
 


midudev:
💡 The Dockerfile seems to be the same between `latest` and `14` or, at least, I don't find any differences.

https://hub.docker.com/layers/keymetrics/pm2/14-alpine/images/sha256-c7545ce31ee5b4a4f8af6ec0731f4928edadf887ca2d1826bd1f3705c3d3277a?context=explore
vs.
https://hub.docker.com/layers/keymetrics/pm2/latest-alpine/images/sha256-6af98b5fffecfac2ca565d3df189980fbc83e18d07359bd81e518f49687f842c?context=explore

PR #892, opened 2020-07-13
Feat/cache decorator add cache key string param

tomasmax:
## Description
Add `cacheKeyString` param in cache decorator, it contains a cache key string(it must be unique). 

It is useful to define a fixed cache key(constructor name + function name, e.g. `cacheKeyString: GetAdListSearchUseCase#execute`) and avoid problems with code minification. 

If By default the following cache key will be created `${target.constructor.name}::${fnName}`

## Related Issue
In milanuncios we detected an issue with minified production code cache keys(`target.constructor.name`)

### DIFF
--- a/packages/sui-decorators/src/decorators/cache/index.js
+++ b/packages/sui-decorators/src/decorators/cache/index.js
@@ -25,7 +25,7 @@ const _cache = ({
 } = {}) => {
   const shouldUseRedisCache = redis && isNode
   const shouldUseMemoryCache = !isNode || !redis
-  const cacheKey = cacheKeyString ?? `${target.constructor.name}::${fnName}`
+  const cacheKey = cacheKeyString || `${target.constructor.name}::${fnName}`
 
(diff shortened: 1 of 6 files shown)


packages/sui-decorators/src/decorators/cache/index.js
```diff
@@ -20,11 +20,12 @@ const _cache = ({
...
   target,
   ttl,
-  redis
+  redis,
+  cacheKeyString
 } = {}) => {
   const shouldUseRedisCache = redis && isNode
   const shouldUseMemoryCache = !isNode || !redis
-  const cacheKey = `${target.constructor.name}::${fnName}`
+  const cacheKey = cacheKeyString ?? `${target.constructor.name}::${fnName}`
```
arnau-rius, line 28: INMHO it is safer to use falsy check (`?`) than using nullish coalescing operator (`??`) in this case.

Issue #895, opened 2020-07-14
[sui-precommit] husky is getting default version of node causing problems


midudev:
As latest versions of node are not supporting mixing ESModules and CommonJS, this is producing some problems when executing some tests. The problem is within husky version, as it's getting the default version instead the one meant to be used with the project. Even the presence of a `.npmrc` file seems to be not working as expected.

PR #896, opened 2020-07-14
Feature/upgrade precommit to use latest husky version

midudev:
## Description
- [x] Upgrade husky version to latest to fix problem that's getting the default node version.
- [x] 🆕 feature: Now precommit is installed automatically on install.
- [x] 🆕 feature: Now on `install` migrate the old config to the new one.
- [x] ✍️ docs: some fixes to the README and documentation

## Related Issue
https://github.com/SUI-Components/sui/issues/895

## Example
With `nvm` using as default `14` but in the session using `10` us causing this:
```
Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /Users/francisco.ruiz/Code/sui/node_modules/@babel/runtime/helpers/esm/asyncToGenerator.js
require() of ES modules is not supported.
require() of /Users/francisco.ruiz/Code/sui/node_modules/@babel/runtime/helpers/esm/asyncToGenerator.js from /Users/francisco.ruiz/Code/sui/packages/sui-decorators/test/server/cacheSpec.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename asyncToGenerator.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /Users/francisco.ruiz/Code/sui/node_modules/@babel/runtime/helpers/esm/package.json.
```

### DIFF
--- a/packages/sui-precommit/package.json
+++ b/packages/sui-precommit/package.json
@@ -1,20 +1,21 @@
 {
   "name": "@s-ui/precommit",
-  "version": "2.7.0",
+  "version": "2.8.0-beta.6",
   "description": "Effortless SUI precommit rules integration in your project",
   "main": "index.js",
   "bin": {
     "sui-precommit": "./bin/sui-precommit.js"
   },
   "scripts": {
-    "postinstall": "sui-precommit install"
+    "postinstall": "node ./bin/sui-precommit.js install"
   },
   "keywords": [],
   "author": "",
   "dependencies": {
-    "@s-ui/lint": "3",
     "@s-ui/helpers": "1",
-    "commander": "5.1.0"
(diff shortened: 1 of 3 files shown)


packages/sui-precommit/package.json
```diff
@@ -6,12 +6,17 @@
...
+  },
   "keywords": [],
   "author": "",
   "dependencies": {
-    "@s-ui/lint": "3",
     "@s-ui/helpers": "1",
-    "commander": "2.9.0"
+    "@s-ui/lint": "3",
+    "commander": "5.1.0",
+    "dlv": "^1.1.3",
```
carlosvillu, line 18: 👀

PR #908, opened 2020-07-29
feat(sui-studio): add setupEnvironment helper to sui-studio

andresin87:
## Description

this pr offerst the posibility to tackle an environment (react-testing-lib environment) for a given component passing its props on each scenario

in [1178](https://github.com/SUI-Components/sui-components/pull/1178) there is a lot of repetitive code in tests for the environment setup

it also creates two default tests to the boilerplate:
```js
it('should render without crashing', () => {
  // Given
  const props = {}

  // When
  const component = <Component {...props} />

  // Then
  const div = document.createElement('div')
  ReactDOM.render(component, div)
  ReactDOM.unmountComponentAtNode(div)
})

it('should NOT render null', () => {
  // Given
  const props = {}

  // When
  const {container} = setup(props)

  // Then
  expect(container.innerHTML).to.be.a('string')
  expect(container.innerHTML).to.not.have.lengthOf(0)
})
```
## Example

```js
/** fooBarComponent Test **/
describe('fooBarComponent', () => {

  const Component = fooBarComponent
  const setup = setupEnvironment(Component)

  it('default context test', () => {
    // given
	const props = {};
	// when
	const {} = setup(props)
	// then
	expect(true).to.be(false)
  })
  
  describe.context.other('fooBarComponent contextified', (fooBarComponentWithOtherContext) => {

	const Component = fooBarComponentWithOtherContext
	const setup = setupEnvironment(Component)

	it('other context test', () => {
	  // given
	  const props = {};
	  // when
	  const {} = setup(props)
	  // then
	  expect(true).to.be(false)
	})
  })
})
```


arnau-rius:
Great 👏

PR #916, opened 2020-08-13
docs(sui-studio): Adding basic styles for Demos standadisation

turolopezsanabria:
**What does this PR do?**
Add basic styles to standardise how the Demos should look like.

**Extra information:**
Nothing will change after publishing these changes, it's just a preparation for the Demos adjustment.

**This is how it looks like:**
styles preview


packages/sui-studio/src/components/layout/_style.scss
```diff
@@ -40,7 +40,146 @@
       transform: translate3d(0, 0, 0);
     }
   }
-
+  .onehalf{
```
jcger, line 43: maybe a whitespace before each `{` ? 😅

turolopezsanabria: 👍 done, I've also removed .onehalf and .onethird, used only for my tests

turolopezsanabria:
> there should be documentation in the readme
Docs in the SUI Studio readme giving visibility to these styles, right?
https://github.com/SUI-Components/sui/tree/master/packages/sui-studio

turolopezsanabria:
Hey, we've just checked and these styles don't get applied in mode Dev. @andresin87 will move the styles to the correct file.

andresin87:
> Hey, we've just checked and these styles don't get applied in mode Dev. @andresin87 will move the styles to the correct file.

done

andresin87 (approved):

PR #927, opened 2020-09-02
feat(sui-test): add support to svg import on null-loader

midudev:
Some 3rd party seems to be trying to load some svg. On sui-test we're adding the possibility to import svg but keeping it as null.

image

### DIFF
--- a/packages/sui-test/package.json
+++ b/packages/sui-test/package.json
@@ -39,14 +39,5 @@
     "mocha": "6.2.2",
     "null-loader": "4.0.0",
     "webpack": "4.41.5"
-  },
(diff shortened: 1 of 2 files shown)


packages/sui-test/package.json
```diff
@@ -39,14 +39,5 @@
     "mocha": "6.2.2",
     "null-loader": "4.0.0",
     "webpack": "4.41.5"
-  },
```
midudev, line 42: This is not needed here, as it's already defined at root level.

kikoruiz (approved):

jcger (approved):

PR #947, opened 2020-10-01
fix(sui-react-router): fixed Routes below a Route that use regexp

### LINKED ISSUE #946: [SUI-ROUTER] When a regex doesn't match, all the routes below doesn't work even though the path matches
### Package

[@s-ui/react-router]@[1.3.0]

### Description

The routes below the one that uses the `regexp` feature, don't work.

### Steps to Reproduce

1. Add a <Route regexp={...}/> (make sure there are other routes below this one)
2. Navigate to one of those routes

**Expected behavior:** The route that has a path that matches the current URL, should be rendered no matter the order of the `<Route>`

**Actual behavior:** The routes that are below the `<Route>` that uses the `regexp` feature, don't work

**Reproduces how often:** (only for bugs) always

devictoribero:
Routes that were placed below a Route with a regexp, were not working

ISSUES CLOSED: #946

## Description
After adding the `regexp` feature to `sui-router`, routes that were placed below that one were not working.

For example:
```
<Route path='/static' component={...} />
<Route regexp={...} component={...} />
<Route path='/static2' component={...} />
```
When visiting `/static` we would have no problem. On the other hand, when visiting `/static2` it would not work.

## Related Issue
fix #946

### DIFF
--- a/packages/sui-react-router/src/internal/createTransitionManager.js
+++ b/packages/sui-react-router/src/internal/createTransitionManager.js
@@ -98,9 +98,8 @@ const createReducerRoutesTree = location => (acc, node) => {
       if (acc.remainingPathname === '') {
         acc.isFinished = checkIntegrity(acc.nodes)
       }
-
-      return acc
     }
+    return acc
   }
 
   if (pattern.charAt(0) === '/') {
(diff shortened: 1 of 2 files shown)


jordevo (approved):

nucliweb (approved):

rmoralp (approved):
Nice! good catch =)

desko27 (approved):

midudev (approved):

kikoruiz (approved):

PR #961, opened 2020-10-19
feat(sui-studio): use native prepare script instead build

midudev:
Use native `prepare` script instead `build` as it will be needed for workspaces.

### DIFF
--- a/packages/sui-studio/bin/sui-studio-generate.js
+++ b/packages/sui-studio/bin/sui-studio-generate.js
@@ -112,7 +112,7 @@ assets`
   "description": "",
   "main": "lib/index.js",
   "scripts": {
-    "build": "npx rimraf ./lib && npm run build:js && npm run build:styles",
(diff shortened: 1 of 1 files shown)


rmoralp (approved):

javiauso (approved):

AgonisticKatai (approved):

andresin87 (approved):

davecarter (approved):

kikoruiz (approved):

jcger:
isn't this a major? what about sui-mono running `npm run build` while releasing?

midudev:
Good call @jcger. Before creating the Pull Request, I checked `sui-mono`. It's checking if the `build` script exists and, if not, just ignore it and try to release the package.

`prepare` script will be automatically executed before releasing the package (native lifecycle of npm). I think this is more correct and it fixes some troubles with workspaces.

jcger (approved):

PR #967, opened 2020-10-27
feat(sui-ci): use fetch instead add huge octokit pkg for calling just one api

midudev:
In order to improve sui-ci installation package time we should remove all possible dependencies not needed.

- [x] Remove not used `execa` dependency.
- [x] Avoid using octokit as it's pretty big and use directly a fetch.


jcger (approved):

andresin87 (approved):

andresin87:
👏

PR #971, opened 2020-10-30
feat(sui-test): add a config to force transpilation of modules via config

midudev:
🆕 Config to add to the package.json the modules that need to be transpiled for `sui-test server`.

This will simplify and fix all the problems we're facing on different projects where some external modules need to be transpilated because are using ESModules.


ruben-martin-lozano (approved):
👌

rmoralp (approved):

PR #1002, opened 2020-11-27
fix(sui-hoc): define default state with prop

nucliweb:
## Description

We use this HOC to add Toggle behavior in components, such as `MoleculeAutosuggest`, and the default state to `isOpen` is **false**, ignoring the prop `isOpen`.

We have a use case to load the Autosuggest showing the options, so we need to use the `isOpen` prop to do it.

## Demo

Contacts Labels

### DIFF
--- a/packages/sui-hoc/src/withOpenToggle.js
+++ b/packages/sui-hoc/src/withOpenToggle.js
@@ -5,7 +5,7 @@ export default BaseComponent => {
 
   return class WithOpenToggle extends Component {
     state = {
-      isOpen: !!this.props.isOpen // eslint-disable-line
+      isOpen: !!this.props.isOpen // eslint-disable-line react/prop-types
     }
 
     static displayName = `WithOpenToggle(${displayName})`
--- a/packages/sui-hoc/src/withOpenToggle.js
+++ b/packages/sui-hoc/src/withOpenToggle.js
@@ -1,14 +1,20 @@
 import React, {Component} from 'react'
+import PropTypes from 'prop-types'
 
 export default BaseComponent => {
   const displayName = BaseComponent.displayName
 
   return class WithOpenToggle extends Component {
     state = {
-      isOpen: !!this.props.isOpen // eslint-disable-line react/prop-types
+      isOpen: !!this.props.isOpen
     }
 
     static displayName = `WithOpenToggle(${displayName})`
+
+    static propTypes = {
+      /** isOpen */
+      isOpen: PropTypes.boolean
+    }
 
     handleToggle = (_, {isOpen} = {}) => {
       this.setState(prevState => ({
--- a/packages/sui-hoc/src/withOpenToggle.js
+++ b/packages/sui-hoc/src/withOpenToggle.js
@@ -5,7 +5,7 @@ export default BaseComponent => {
 
   return class WithOpenToggle extends Component {
     state = {
-      isOpen: Boolean(this.props.isOpen) // eslint-disable-line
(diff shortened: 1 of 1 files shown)


javiauso (approved):

packages/sui-hoc/src/withOpenToggle.js
```diff
@@ -5,7 +5,7 @@ export default BaseComponent => {
 
   return class WithOpenToggle extends Component {
     state = {
-      isOpen: false
+      isOpen: Boolean(this.props.isOpen) // eslint-disable-line
```
desko27, line 8: Why the eslint comment? 😮 Anyway, could `!!` be used instead?

```suggestion
      isOpen: !!this.props.isOpen
```

desko27 (approved):

packages/sui-hoc/src/withOpenToggle.js
nucliweb: Screen Shot 2020-11-27 at 13 28 24

kikoruiz: You can do something more specific: `// eslint-disable-line react/prop-types`

quinwacca (approved):

ivanmlaborda (approved):

packages/sui-hoc/src/withOpenToggle.js
```diff
@@ -5,7 +5,7 @@ export default BaseComponent => {
 
   return class WithOpenToggle extends Component {
     state = {
-      isOpen: false
+      isOpen: !!this.props.isOpen // eslint-disable-line
```
jordevo, line 8: 🍆 do you still need to disable this line on eslint?

jordevo (approved):

packages/sui-hoc/src/withOpenToggle.js
nucliweb: 🤪  👉🏼   https://github.com/SUI-Components/sui/pull/1002#discussion_r531574376

packages/sui-hoc/src/withOpenToggle.js
midudev: The bundle is not affected by adding PropTypes as we're removing them for Pro bundles.

packages/sui-hoc/src/withOpenToggle.js
jordevo: sorry, hadn't seen the resolved convo! ;)

packages/sui-hoc/src/withOpenToggle.js
nucliweb: @midudev @desko27 @kikoruiz PropTypes added and I removed the eslint comment

packages/sui-hoc/src/withOpenToggle.js
```diff
@@ -1,15 +1,21 @@
 import React, {Component} from 'react'
+import PropTypes from 'prop-types'
 
 export default BaseComponent => {
   const displayName = BaseComponent.displayName
 
   return class WithOpenToggle extends Component {
     state = {
-      isOpen: false
+      isOpen: !!this.props.isOpen
```
andresin87, line 9: https://medium.com/dev-genius/vs-boolean-cd502fbd6101

I personally prefer Boolean(whatever)

desko27: I suggested `!!` because IMHO looks shorter and equally understandable for non-complex expressions, but as the article states (good one BTW) both are nice options!

jcger (approved):

PR #1006, opened 2020-12-09
feat(sui-react-initial-props): add possibility to keep mounted the component

midudev:
**Solution was provided by @desko27 with some minor changes.**

- [x] Add new `keepMounted` property to be added to the page in order to avoid component unmount.
- [x] `isLoading` prop is provided to component if `keepMounted` is used while getting initial props.
- [x] Avoid the need of having to use contextFactory for client and get from the context.
- [x] Use new jsx runtime.


desko27 (approved):
🙌

jcger (approved):
sounds familiar 👏

kikoruiz (approved):

rmoralp (approved):

packages/sui-react-initial-props/src/withInitialProps.js
```diff
@@ -1,37 +1,110 @@
...
-          setState({initialProps: {error}, isLoading: false})
-        })
-    }, [])
+    useEffect(() => {
+      // check if got initial props from window, because then there's no need
+      // to request them again from client
+      if (initialPropsFromWindowRef.current) {
+        initialPropsFromWindowRef.current = undefined
+      } else {
+        Page.getInitialProps({context, routeInfo})
```
rmoralp, line 62: We moved that useEffect from an only one render when component mounts to a effect that is going to be executed every time `routeInfo.location` changes.
In this case, I think we need to set `isLoading` to `true` when it happens, just before we resolve `getInitialProps` with updated `routeInfo`.

```suggestion
        setState({initialProps, isLoading: true})

        Page.getInitialProps({context, routeInfo})
```

packages/sui-react-initial-props/src/withInitialProps.js
```diff
@@ -1,37 +1,112 @@
...
+      isLoading: !initialPropsFromWindowRef.current
+    }))
+
+    useEffect(() => {
+      // check if got initial props from window, because then there's no need
+      // to request them again from client
+      if (initialPropsFromWindowRef.current) {
+        initialPropsFromWindowRef.current = undefined
+      } else {
+        setState({initialProps, isLoading: true})
```
desko27, line 62: We should even check `requestedInitialPropsOnceRef.current` since when false, then `isLoading` is already true, so no need to update state.

```js
if (requestedInitialPropsOnceRef.current) {
  setState({initialProps, isLoading: true})
}
```

rmoralp:
Beta live en producción https://motos.coches.net/segunda-mano/

packages/sui-react-initial-props/src/withInitialProps.js
midudev: Lo reviso!

packages/sui-react-initial-props/src/withInitialProps.js
```diff
@@ -1,37 +1,112 @@
...
         : null
     }
 
-    return isLoading ? renderLoading() : <Page {...initialProps} {...props} />
+    return isLoading ? renderLoading() : renderPage()
   }
+
+  // if `keepMounted` property is found and the component is the same one,
+  // we just reuse it instead of returning a new one
+  if (Page.keepMounted && Page.name === latestClientPage?.Page?.name) {
```
rmoralp, line 100: Shouldn't we use displayName instead name?
```suggestion
  if (Page.keepMounted && Page.displayName === latestClientPage?.Page?.displayName) {
```

rmoralp:
Also in production for 🚗   in https://www.coches.net/km-0/

PR #1014, opened 2020-12-11
fix(sui-mono): avoid sui-mono to get wrong folders

midudev:
**Fixes**
- [x] Avoid getting `node_modules` folders as it doesn't make sense and is causing the `undefined` weird package.

For example `/demo/node_modules/any-package/folder` was used as scope if `demo` root is installing packages.

- [x] Avoid getting folders that don't have `package.json` as a `npm install` will cause to reinstall dependencies for first parent folder with `package.json`. 

This will avoid executing an unnecessary `npm install` on `/demo/**/**/themes` folder as `package.json` doesn't exist and also will skip `demos`  with the `package.json` file. For example, Fotocasa Components goes from 55 `npm install` for demos to `10` and get rid of all the errors.

Error on Travis CI Enterprise we're looking to remove: https://travis.mpi-internal.com/scmspain/frontend-mt--uilib-components/builds/3768262#L306

**Refactors**
- [x] Simplify reading config object by using destructuring instead a method.
- [x] Simplify how to get the scopes and explain every line with an example.
- [x] Avoid using `flatten` created method.
- [x] Rename internal `cwds` method to `getDeepFolders` for clarity.
- [x] Create easily an Array of n items without the dark magic.
- [x] Reuse `getPackageJson` from helpers.

The dark magic 🧙‍♂️:
```js
const baseFolders = Array.apply(null, Array(deep)).map(
    Number.prototype.valueOf,	
    0	
)
```

The easiest way 👍:
```js
[...Array(deep)]
```

### DIFF
--- a/packages/sui-mono/src/config.js
+++ b/packages/sui-mono/src/config.js
@@ -1,71 +1,55 @@
 const path = require('path')
-const readdirSync = require('fs').readdirSync
-const statSync = require('fs').statSync
+const {existsSync, readdirSync, statSync} = require('fs')
+const {getPackageJson} = require('@s-ui/helpers/packages')
 
-const basePath = process.cwd()
-const projectPackage = require(path.join(basePath, 'package.json'))
-const packageConfig = projectPackage.config
+const CWD = process.cwd()
 const ROOT_SCOPE = 'Root'
 
-function getOrDefault(key, defaultValue) {
-  return (
-    (packageConfig &&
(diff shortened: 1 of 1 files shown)


jcger (approved):

adriadoz (approved):

andresz1 (approved):

andresin87:
👏👏👏

PR #1023, opened 2020-12-18
feat(sui-component-dependencies): add deprecate proptype function

andresin87:
## Description

For deprecating a component prop. It only warns if NODE_ENV !== 'production' and a prop has a given value.

## Example

```js
// Component.js
import Proptypes from 'prop-types'
import deprecate from '@s-ui/component-dependencies/deprecate'

...
const Component = ({type}) => <div>{type}</div>

Component.propTypes = {
  type: deprecate(PropTypes.number)
  customMessageDeprecatedProp: deprecate(PropTypes.number, (props, propName, componentName) => {
  	console.warn(`prop {propName} is deprecated in {componentName}`)
  })
}
...

export default Component
```
```js
// main.js
import {render} from 'react-dom'
import Component from './Component'

render(element, <Component type={1}/>)
```

Will output by default:

image


andresin87:
up ⬆

PR #1024, opened 2020-12-21
feat(sui-domain): subscribe to domain Use Case without decorators

midudev:
- [x] Remove the need of a manual decorator to enable this feature. 🧽
- [x] Improve domain predictability. 💭 
- [x] Avoid having to download the chunk for subscribing to the useCase! ⚡
- [x] Allow a simpler API surface for the domain: 🏄‍♂️

```
domain.get('generate_search_url_search_use_case').execute...
domain.get('generate_search_url_search_use_case').subscribe...
```

How to use:
```js
domain.get('generate_search_url_search_use_case').subscribe(({params, error, result}) => {
  // doSomething
})
```

How to unsubscribe:
```js
const unsubscribe = domain.get('generate_search_url_search_use_case').subscribe(handleSubscription)
unsubscribe()
```

Unsubscribe on React useEffect:
```js
useEffect(() => {
    const {unsubscribe} = domain.get('generate_search_url_search_use_case').subscribe(handleSubscription)
    return unsubscribe
  }, [deps])
```


desko27:
😮😮😮 This is the way

desko27 (approved):

adriadoz (approved):

javiauso (approved):

rmoralp (approved):

packages/sui-domain/src/EntryPointFactory.js
```diff
@@ -42,13 +43,34 @@ export default ({useCases, config}) =>
...
+                  )
+                  return Promise.reject(e)
+                })
+            },
+            subscribe: callback => {
+              this.subscribers[key] = this.subscribers[key] || []
+              this.subscribers[key].push(callback)
+
+              // return a way to remove the listener
+              return () => this.subscribers[key].filter(l => l !== callback)
```
rmoralp, line 73: As a subjective opinion, I like more an `rxjs` API where you can get a subscription with a `unsubscribe` method.
Getting directly the `unsubscribe` function when `subscribe` has been executed can cause confusion to me.
Some `rxjs` docs: https://rxjs-dev.firebaseapp.com/guide/subscription

rmoralp (approved):
Great job!

packages/sui-domain/src/EntryPointFactory.js
midudev: Then we could do something like this. The drawback I see is that we will lose the possibility of cleaning the effect without storing a variable.

```
useEffect(() => {
    const {unsubscribe} = domain.get('generate_search_url_search_use_case').subscribe()
// ...
    return unsubscribe
  }, [])
```

oriolpuig (approved):

nucliweb (approved):

andresz1 (approved):

quinwacca (approved):

AgonisticKatai (approved):

nucliweb (approved):

danivm (approved):

kikoruiz (approved):
💚

packages/sui-domain/README.md
```diff
@@ -101,3 +101,17 @@ export default class CurrentUserService extends Service {
   }
 }
 ```
+
+## Listen a useCase
+
+`@s-ui/domain` includes a way to subscribe to eveyr useCase execution without the need of using any kind of decorator or external dependency.
```
jcger, line 107: typo every

jcger (approved):

packages/sui-domain/README.md
midudev: Thanks!

PR #1029, opened 2021-01-08
feat(sui-studio): add test timeout option

andresz1:
## Description
This PR adds a way to set a `timeout` for the sui browser tests

## Related Issue
None

## Example
```
sui-studio test --timeout 10000
```

### DIFF
--- a/packages/sui-studio/bin/sui-studio-test.js
+++ b/packages/sui-studio/bin/sui-studio-test.js
@@ -38,7 +38,7 @@ const run = async () => {
           ),
           program.watch && '--watch',
           program.ci && '--ci',
-          program.timeout !== undefined && `-T ${program.timeout}`
+          program.timeout && `-T ${program.timeout}`
         ].filter(Boolean),
         {
           shell: false,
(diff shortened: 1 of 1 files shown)


javiauso (approved):

emiliovillu (approved):

jmanrumartinez (approved):

packages/sui-studio/bin/sui-studio-test.js
```diff
@@ -36,7 +37,8 @@ const run = async () => {
             'index.js'
           ),
           program.watch && '--watch',
-          program.ci && '--ci'
+          program.ci && '--ci',
+          `-T ${program.timeout}`
```
rmoralp, line 41: I suggest to make it conditional and avoid default value in order to not suppose that our runtime uses always 2 seconds
```
          program.timeout && `-T ${program.timeout}`
```

rmoralp (approved):

packages/sui-studio/bin/sui-studio-test.js
andresz1: Agree!

PR #1040, opened 2021-02-01
Feat/add workspaces to sui mono

midudev:
- [x] Integrate @s-ui/cz in @s-ui/mono
- [x] Remove support for `deepLevel` and `packagesFolder` and use `workspaces` instead.
- [x] Rely only on `prepublish` or `prepare` instead `build` script.
- [x] Remove support for Node 10 and older.


packages/sui-mono/bin/sui-mono-changelog.js
```diff
@@ -30,21 +28,20 @@ program
   .parse(process.argv)
 
 const CHANGELOG_NAME = getChangelogFilename()
-const cwd = path.join(process.cwd(), getPackagesFolder())
-const getRepoFolders = () =>
-  !isMonoPackage() ? getPackagesPaths(cwd)(getScopes()) : ['.']
-const folders = program.args.length ? program.args : getRepoFolders()
+
+const folders = program.args.length ? program.args : getWorkspaces()
```
jcger, line 32: do you know where we are using this program.args use case?

jcger (approved):
would be nice if we could try to add the first tests to it

packages/sui-mono/bin/sui-mono-changelog.js
midudev: No idea, tbh.

rmoralp (approved):

jcger (approved):

kikoruiz (approved):

PR #1047, opened 2021-02-11
Sui studio utils/mock per chunks

jcger:
## Description
Easier way to add multiple use cases at once

## Example

```js
import { DomainBuilder } from '@s-ui/studio-tools'
import myDomain from 'domain'
const domain = DomainBuilder.extend({ myDomain })
  .mockUseCases([
    ['get_products_use_case', {success: ['pineapple', 'apple', 'strawberry', 'coffee']}],
    ['get_user_use_case', {fail: {code: 500, data: {}}}]
  ]).build()
```

### DIFF
--- a/packages/sui-studio-utils/README.md
+++ b/packages/sui-studio-utils/README.md
@@ -44,6 +44,25 @@ domain.get('current_user_use_case').execute().then((products) => {
 })
 ```
 
+#### more than one use case at a time
+
+```js
+import { DomainBuilder } from '@s-ui/studio-tools'
+import myDomain from 'domain'
+
+const domain = DomainBuilder.extend({ myDomain })
+  .mockUseCases([
+    ['get_products_use_case', respondWith: {success: ['pineapple', 'apple', 'strawberry', 'coffee']}],
+    ['get_user_use_case', respondWith: {fail: {code: 500, data: {}}}]
+  ]).build()
+
+// Execute the use case and check if everything works
+domain.get('current_user_use_case').execute().then((products) => {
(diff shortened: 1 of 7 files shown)


kikoruiz (approved):

packages/sui-studio-utils/README.md
```diff
@@ -44,6 +44,25 @@ domain.get('current_user_use_case').execute().then((products) => {
...
 ```
 
+#### more than one use case at a time
+
+```js
+import { DomainBuilder } from '@s-ui/studio-tools'
+import myDomain from 'domain'
+
+const domain = DomainBuilder.extend({ myDomain })
+  .mockUseCases([
```
ferransimon, line 54: What about calling multiple time to the function `for()` instead of having 2 different methods of adding mocks?
Could it be something like
`
DomainBuilder.extend({ myDomain }).for({useCase: 'get_products_use_case'}).respondWith(response).and().for({useCase: 'get_user_use_case'}).respondWith(userResponse).build()
`

tonyghiani (approved):

packages/sui-studio-utils/src/domain-builder/index.js
```diff
@@ -10,6 +10,13 @@ export default class DomainBuilder {
     this._useCases = {}
   }
 
+  mockUseCases(mocks) {
+    mocks.forEach(
```
tonyghiani, line 14: You may want to re-use the existing methods to exploit the validation logic they already implement, like the respondWith method validation for success and fail:
```
mocks.forEach(([useCase, responses]) =>
  this.for({useCase}).respondWith(responses)
)
```

jcger:
it wasn't needed

PR #1060, opened 2021-02-26
fix(packages/sui-studio): make test folder really optional

midudev:
Fix making test folder really optional

### DIFF
--- a/packages/sui-studio/src/runtime-mocha/index.js
+++ b/packages/sui-studio/src/runtime-mocha/index.js
@@ -15,12 +15,18 @@ const importAll = request => request.keys().forEach(request)
 const originalKarmaLoader = window.__karma__.loaded
 window.__karma__.loaded = () => {}
 
-// get all tests files available using a regex
-const testsFiles = require.context(
-  `${__BASE_DIR__}/test/`,
-  true,
-  /\.\/(\w+)\/(\w+)\/index.(js|jsx)$/
-)
+// default testFiles in case /test folder does not exist
+let testsFiles = () => {}
+testsFiles.keys = () => []
+
+try {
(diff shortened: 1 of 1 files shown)


kikoruiz (approved):

desko27 (approved):

klaufel (approved):

oriolpuig (approved):

packages/sui-studio/src/runtime-mocha/index.js
```diff
@@ -15,12 +15,18 @@ const importAll = request => request.keys().forEach(request)
...
-const testsFiles = require.context(
-  `${__BASE_DIR__}/test/`,
-  true,
-  /\.\/(\w+)\/(\w+)\/index.(js|jsx)$/
-)
+// default testFiles in case /test folder does not exist
+let testsFiles = () => {}
+testsFiles.keys = () => []
+
+try {
```
juanmendez90, line 22: why you opt to implement a try/catch instead of checking if the `${__BASE_DIR__}/test/` exists?
I think the second option increases the readability of the code

juanmendez90 (approved):

packages/sui-studio/src/runtime-mocha/index.js
midudev: While this seems to be running on node, this is browser code but bundled by Webpack.

`require.context` is a Webpack internal method that allows us to extract files from files by using regex queries statically. We can't know dynamically if there's a `test` folder present using Node.

So, the best approach is to allow Webpack to try to use require.context and, if not, just avoid the Error.

We could add a Babel plugin to do the check but it would complicate the code instead.

PR #1067, opened 2021-03-02
fix(packages/sui-test): fix new test path in studio

kikoruiz:
## Description
Fix an existing magic that was making testable components in studio could have their contexts available, but it started to fail since a new path for tests was created in `sui-studio`.

### DIFF
--- a/packages/sui-test/bin/karma/babelPatch.js
+++ b/packages/sui-test/bin/karma/babelPatch.js
@@ -24,15 +24,16 @@ module.exports = function({types: t}) {
           // by reversing the path and exlcuding the first slash (it's /index.js)
           const reversedPath = state.file.opts.filename.split('/').reverse()
           const [, lastPath] = reversedPath
-          const isNewPath = lastPath === 'test'
+          const isRelativeTestPath = lastPath === 'test'
           let componentKey = ''
-          if (isNewPath) {
+          if (isRelativeTestPath) {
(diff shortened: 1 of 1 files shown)


midudev (approved):

rmoralp (approved):

andresin87 (approved):

packages/sui-test/bin/karma/babelPatch.js
```diff
@@ -25,12 +22,17 @@ module.exports = function({types: t}) {
...
-            .reverse()
-          // we create a key based on the category and component path
-          const componentKey = `${category}/${component}`
-
+          const reversedPath = state.file.opts.filename.split('/').reverse()
+          const [, lastPath] = reversedPath
+          const isNewPath = lastPath === 'test'
+          let componentKey = ''
+          if (isNewPath) {
+            const [, , component, category] = reversedPath
```
andresin87, line 30: Maybe a better naming like isRelativeTestDirectoryPath more declarative. When do 'newPath' name becomes oldPath? 🥸

andresin87: A newer path available will become newNewPath or maybe theNewestPath

kikoruiz: It's fine, but the idea is to kill one of the paths (the old one).

PR #1077, opened 2021-03-10
chore(packages/sui-pde): react as peer dependency

jcger:
## Description
sui-pde exports hooks which need react and therefore need it as dependency although the exact version should be set by the package which uses those hooks


package.json
```diff
@@ -24,6 +24,9 @@
     "test:server:watch": "npm run test:server -- --watch",
     "test:e2e": "node ./packages/sui-studio/test/server/integration/static-server.js ./packages/sui-studio/test/server/integration/sample-studio/public && npx sui-test e2e --baseUrl=http://localhost:1234"
   },
+  "dependencies": {
```
midudev, line 27: Pero esta ya no es necesaria, no?

jcger: yep, you're right

midudev (approved):

PR #1078, opened 2021-03-11
Feature/add user consents hook

desko27:
## Description
- Add React hook that returns whether a user has given its consent to a set of consents or not, and updates over time if the user changes the consents.
- Add server and browser tests.

```js
import {useUserConsents} from '@s-ui/consents'

// will return true if user did accept at least to consents 1, 2 and 3
const isAccepted = useUserConsents([1, 2, 3])
```


ruben-martin-lozano (approved):

adriadoz (approved):
Just what I needed :D

packages/sui-consents/src/useUserConsents.js
```diff
@@ -0,0 +1,46 @@
...
+  const [areConsentsAccepted, setAreConsentsAccepted] = useState(() =>
+    hasUserConsents({requiredConsents, cookies})
+  )
+
+  /**
+   * From then on, we listen for TCF events so consents changes
+   * can be catched right away, no matter when, and react to them
+   * wherever the hook is used.
+   */
+  useEffect(() => {
```
midudev, line 27: Maybe not a big thing but this is going to add a new event listener for each hook call.

packages/sui-consents/src/useUserConsents.js
```diff
@@ -0,0 +1,46 @@
...
+ * @param {array} requiredConsents for example [1, 3, 8]
+ * @return {boolean}
+ */
+export default function useUserConsents(requiredConsents) {
+  /**
+   * Consents acceptance state is inited based on the cookies from the
+   * context, so we know the state of consents from the beginning, even
+   * in SSR.
+   */
+  const {cookies} = useContext(SUIContext)
```
midudev, line 17: I'm not sure `cookies` is available from `SUIContext` on all verticals. 

At least, we should mention this particularity on the README to make this work.

packages/sui-consents/src/useUserConsents.js
desko27: That's right. I don't think it's a problem since consents doesn't change a lot while using the app (mostly never). Considering this, I don't if there would be any gain by setting up a context or similar for this. Any other thoughts on this are welcome.

midudev: Not a big deal, let's keep it in mind in case we see some degradation about this.

adriadoz: As @desko27 says consents don't change so often. So imho we could just go ahead and if in a near future if it is a problem we can add the context.

packages/sui-consents/src/useUserConsents.js
desko27: Done 👌🏻

paulusrex (approved):

adriadoz (approved):

packages/sui-consents/README.md
```diff
@@ -26,3 +26,18 @@ hasUserConsents({
   cookies
 })
 ```
+
+## useUserConsents
+
+React hook that returns whether a user has given its consent to a set of consents or not, and updates over time if the user changes the consents.
+
+**IMPORTANT NOTE:** This hook **requires** an isomorphic **`cookies`** property to reside **in your SUIContext**. This is necessary for the initial state to work.
```
midudev, line 34: 👏

midudev (approved):

kikoruiz (approved):

PR #1080, opened 2021-03-15
Feat/launch experiment viewed

jcger:
in order to be able to replicate the experiment in our analytics tool we need to send the Experiment Viewed event. This event will only be launched when the adapter returned a variation (aka optimizely already launched its impression event)

As the experiment viewed should only be launched from client side, I implemented different strategies for server/client, making server side stop activating an experiment but just getting the variation

### DIFF
--- a/packages/sui-pde/src/hooks/useExperiment.js
+++ b/packages/sui-pde/src/hooks/useExperiment.js
@@ -18,11 +18,7 @@ export default function useExperiment(experimentName, attributes) {
     try {
       variationName = pde.activateExperiment({name: experimentName, attributes})
 
-      if (
-        variationName &&
-        typeof window !== 'undefined' &&
-        window.analytics?.track
-      ) {
+      if (variationName && window.analytics?.track) {
         window.analytics.track('Experiment Viewed', {
           experimentName,
           variationName
--- a/packages/sui-pde/src/hooks/useExperiment.js
+++ b/packages/sui-pde/src/hooks/useExperiment.js
@@ -17,6 +17,21 @@ export default function useExperiment(experimentName, attributes) {
 
     try {
       variationName = pde.activateExperiment({name: experimentName, attributes})
+
+      if (
+        variationName &&
+        typeof window !== 'undefined' &&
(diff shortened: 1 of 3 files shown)


packages/sui-pde/test/common/useExperimentSpec.js
```diff
@@ -1,10 +1,21 @@
+/* eslint-disable no-console */
 import {cleanup, renderHook} from '@testing-library/react-hooks'
 import {expect} from 'chai'
 import PdeContext from '../../src/contexts/PdeContext'
 import useExperiment from '../../src/hooks/useExperiment'
 import sinon from 'sinon'
 
-describe('useExperiment hook', () => {
+describe.client('useExperiment hook', () => {
```
jcger, line 8: gonna rollback this

kikoruiz (approved):

packages/sui-pde/src/hooks/useExperiment.js
```diff
@@ -17,6 +17,17 @@ export default function useExperiment(experimentName, attributes) {
 
     try {
       variationName = pde.activateExperiment({name: experimentName, attributes})
+
+      if (variationName && window.analytics?.track) {
+        window.analytics.track('Experiment Viewed', {
```
rmoralp, line 22: Does it have the sense to use `window.sui.analytics` in order to get the segment wrapper benefits?
May could be a way to pass a custom callback or sth else to make this track? I mean, doing `window.analytics.track` I think that we are making this package less "open source"

jcger: sui.analytics shouldn't be an issue as the segment wrapper does patch window.analytics https://github.mpi-internal.com/scmspain/frontend-all--segment-wrapper/blob/master/src/patchAnalytics.js

regarding the callback question, yep you are right although I personally prefer adding it once someone requests it instead of having code that actually no one uses. Maybe it's something to discuss within the team, what do we wan't to do in these cases? What should be our approach from now on?

rmoralp (approved):

PR #1081, opened 2021-03-16
Feat/setup environment rtl api

rmoralp:
## Description
setupEnvironment does not have enabled React Testing Library render options.
Also, the second parameter is not needed because sui-studio injects SUI context to the component and no one is using it.

Developers needs to have enabled RTL render options.
Also, in order not to create a wrapper with multiple context providers and duplicate that code across all test files, we implement a `contexts` option to do it inside setupEnvironment if dev wants. Of course it could be done using only `wrapper` render option.

RTL docs: https://testing-library.com/docs/react-testing-library/api/#render-options
RTL Hooks context docs: https://react-hooks-testing-library.com/usage/advanced-hooks#context

## Example
As an example, a component using SUI Context and other context (sui-react-router context):
```
const setup = setupEnvironment(Component, {
      contexts: [{provider: RouterContext.Provider, value: router}]
    })
```

### DIFF
--- a/packages/sui-studio/src/environment-mocha/setupEnvironment.js
+++ b/packages/sui-studio/src/environment-mocha/setupEnvironment.js
@@ -1,21 +1,54 @@
-import SUIContext from '@s-ui/react-context'
+import PropTypes from 'prop-types'
 import {render} from '@testing-library/react'
 
-const setupEnvironment = (Component, context) => (props, ref) => {
-  const container = document.createElement('div')
-  container.setAttribute('id', 'test-container')
-  return render(
-    context ? (
-      <SUIContext.Provider value={context}>
-        <Component {...props} ref={ref} />
-      </SUIContext.Provider>
-    ) : (
-      <Component {...props} ref={ref} />
-    ),
-    {
-      container: document.body.appendChild(container)
-    }
+const DefaultWrapper = ({children}) => <>{children}</>
+
+DefaultWrapper.propTypes = {
+  children: PropTypes.any
+}
+
+const getWrapper = ({contexts = [], wrapper: Wrapper = DefaultWrapper}) => {
+  const ContextsWrapper = ({children}) =>
+    contexts.reduce(
+      (children, {provider: Provider, value}) => (
+        <Provider value={value}>{children}</Provider>
+      ),
+      children
+    )
+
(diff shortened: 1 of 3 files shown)


packages/sui-studio/src/environment-mocha/setupEnvironment.js
```diff
@@ -1,21 +1,54 @@
-import SUIContext from '@s-ui/react-context'
+import PropTypes from 'prop-types'
 import {render} from '@testing-library/react'
 
-const setupEnvironment = (Component, context) => (props, ref) => {
```
rmoralp, line 4: ℹ️  no one uses that second parameter

andresin87 (approved):

midudev (approved):

jcger (approved):

javiauso (approved):

arnau-rius:
Since we are changing the API maybe it is interesting to implement the same approach  that appears in the documentation: 
```js
const customRender = (ui, options) =>
  render(ui, { wrapper: AllTheProviders, ...options })
```
https://testing-library.com/docs/react-testing-library/setup/

arnau-rius (approved):

packages/sui-studio/src/environment-mocha/setupEnvironment.js
```diff
@@ -1,21 +1,54 @@
...
+}
+
+/**
+ * Render a component using React Testing Library (RTL)
+ * Keep RTL api
+ * https://testing-library.com/docs/react-testing-library/api/#render-options
+ */
+const setupEnvironment = (
+  Component,
+  {contexts = [], hydrate, queries, wrapper} = {}
```
andresin87, line 39: `{contexts = [], hydrate, queries, wrapper, ...otherSettings} = {}`

add otherSettings destructuring in render config (L46)

rmoralp: RTL render method does not have more options than used on L46
Well, it has `baseElement` but if the `container` options is specified, then this defaults to that.
See: https://testing-library.com/docs/react-testing-library/api/#render-options

kikoruiz (approved):

jordevo (approved):

PR #1088, opened 2021-03-19
feat(packages/sui-test): add headless and default command timeout options

desko27:
## Description

Adds some options for e2e command:

- **headless**: if you set a different browser by using `--browser` option, can't be headless unless you specify it.
- **defaultCommandTimeout**: since default is 4s, we need to set a higher value in order to ensure some tests pass in CI, because CI processes are sometimes slower than local.

### DIFF
--- a/packages/sui-test/README.md
+++ b/packages/sui-test/README.md
@@ -155,6 +155,7 @@ module.exports = (on, config) => {
   Options:
 
     -B, --baseUrl <baseUrl>                  URL of the site to execute tests (in ./test/e2e/) on.
+    -dct, --defaultCommandTimeout <ms>       Time, in milliseconds, to wait until most DOM based commands are considered timed out.
     -S, --screenshotsOnError                 Take screenshots of page on any failure.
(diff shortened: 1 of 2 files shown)


rmoralp (approved):

packages/sui-test/README.md
```diff
@@ -155,13 +155,15 @@ module.exports = (on, config) => {
   Options:
 
     -B, --baseUrl <baseUrl>                  URL of the site to execute tests (in ./test/e2e/) on.
+    -dct, --defaultCommandTimeout <ms>       Time, in milliseconds, to wait until most DOM based commands are considered timed out.
```
midudev, line 158: igual en lugar de `dct` podría ser -T en mayúscula, no?

midudev (approved):

packages/sui-test/README.md
desko27: 👌🏻 Pues sí, ¡mejor!

danivm (approved):

PR #1093, opened 2021-03-23
feat(sui-ssr): check criticalCss when empty string

javiauso:
## Description
sui-ssr: check criticalCss when empty string


igncalleja (approved):

rmoralp:
It is not the bug we are try to look for. I'll explain you more in detail in a meeting

PR #1100, opened 2021-03-31
feat(packages/sui-ssr): inline dynamic import styles on demand

rmoralp:
## Description
- **`createStylesFor`** (`undefined`): Define how the server should manage style imports. When separate styles per page, by default those imports will load css asynchronously but maybe we want that server to add them as an `<link>` in app head and be async or not depends on criticalCSS.

  - **`appStyles`** (`string`): Define the webpackChunkName of app styles. These styles usually are imported in `app.js`.
  - **`createPagesStyles`** (`false`): Define if pages have style imports that should be managed by the server. The server will get these files from the `asset-manifest.json` file, which should be created by sui-ssr.

> ⚠️ `createStylesFor` does not works as expected when sui-bundler configuration has `onlyHash` defined to `true`.

Published as a beta: `7.21.0-beta.2`

This code is based on FC custom server and has been tested in MA with beta version.

## Example
```
"sui-ssr": {
      "createStylesFor": {
        "appStyles": "AppStyles",
        "createPagesStyles": true
      }
    }
```

### DIFF
--- a/packages/sui-ssr/README.md
+++ b/packages/sui-ssr/README.md
@@ -277,6 +277,13 @@ Configs accepted:
 
 - **`serverContentType`** (`undefined`): A valid Content-Type string to be set in response header Content-Type. If not defined, it will use the regular html type with utf-8 charset encoding.
 
+- **`createStylesFor`** (`undefined`): Define how the server should manage style imports. When separate styles per page, by default those imports will load css asynchronously but maybe we want that server to add them as an `<link>` in app head and be async or not depends on criticalCSS.
+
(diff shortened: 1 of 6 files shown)


javiauso (approved):

jcger (approved):

andresz1 (approved):

AgonisticKatai (approved):

kikoruiz (approved):
Nice feature!! 👏

adriadoz (approved):

PR #1103, opened 2021-04-08
feat(packages/sui-pde): global attributes to be always used

jcger:
## Description
Adds a way to add global attributes so they are set up once and reused every time. The idea behind this PR is to send the site and environment on every impression event

## Example

```js
const optimizelyAdapter = new OptimizelyAdapter({
  optimizely: optimizelyInstance, // created before
  userId: '1234',
  hasUserConsents: true,
  applicationAttributes: {   // optional, global application attributes that must be send on every experiment activation
    site: 'coches.net',
    environment: 'development'
  }
})
```


kikoruiz (approved):
Great!

rmoralp (approved):

PR #1111, opened 2021-04-21
Fix/async styles critical remove

rmoralp:
## Description
Do not remove critical CSS `<style>` when CSS files `onload`
If our app has more than one CSS file, the first file loaded will remove the critical CSS node and we will see a FOUC.

Captura de pantalla 2021-04-20 a las 8 01 41

Bonus:
use media `print` instead `only x` because of when browsers encounter media types that don’t match, they currently treat them all the same - they load the file anyway.
See: https://www.filamentgroup.com/lab/load-css-simpler/#why-not-use-a-faux-media-attribute%3F

Beta published 👉 @s-ui/ssr@7.23.0-beta.2

### DIFF
--- a/packages/sui-ssr/server/config.js
+++ b/packages/sui-ssr/server/config.js
@@ -7,7 +7,7 @@ const DEFAULT_VALUES = {
 }
 
 const ASYNC_CSS_ATTRS =
-  'rel="stylesheet" media="only x" as="style" onload="this.media=\'all\';"'
+  'rel="stylesheet" media="print" as="style" onload="this.media=\'all\'"'
 
 let ssrConfig
 try {
--- a/packages/sui-ssr/server/config.js
+++ b/packages/sui-ssr/server/config.js
@@ -7,7 +7,7 @@ const DEFAULT_VALUES = {
 }
 
 const ASYNC_CSS_ATTRS =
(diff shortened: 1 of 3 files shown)


javiauso (approved):

AgonisticKatai (approved):

packages/sui-ssr/server/config.js
```diff
@@ -7,7 +7,7 @@ const DEFAULT_VALUES = {
 }
 
 const ASYNC_CSS_ATTRS =
-  'rel="stylesheet" media="only x" as="style" onload="this.media=\'all\';var e=document.getElementById(\'critical\');if(e){e.parentNode.removeChild(e);}"'
+  'rel="stylesheet" media="print" as="style" onload="this.media=\'all\'"'
```
kikoruiz, line 10: I think the idea is to leave a non-valid media type... is that a problem?

andresz1 (approved):

packages/sui-ssr/server/config.js
rmoralp: It could be, as I read on filament group page: https://www.filamentgroup.com/lab/load-css-simpler/#why-not-use-a-faux-media-attribute%3F

kikoruiz: Ok, I didn't understand it right then....

kikoruiz (approved):

PR #1112, opened 2021-04-21
fix(packages/sui-test): add an option to prevents compile lib folders…

rmoralp:
… for server tests
## Description
In some projects, if we add `/lib/` folder to the mocha register chunk increases over 500Kb and 
all tests fail.
With option `useLibDir` we could prevent compiling lib folders on the mocha runner if needed

[Failing build](https://travis.mpi-internal.com/scmspain/frontend-ma--lib-milanuncios/builds/4873285)

### DIFF
--- a/packages/sui-test/README.md
+++ b/packages/sui-test/README.md
@@ -203,6 +203,7 @@ If defined, any error on your tests will create a screenshot of that moment in t
 - `server`: Config for `@s-ui/test server` binary:
   - `forceTranspilation`: List of regexs (string based, later will be transformed with `new Regex`) of modules to transpile. This is useful in case you're using server tests for modules that are ESModules based and need to be transpiled with `@babel/plugin-transform-modules-commonjs`.
   - `esmOverride`: Boolean flag (defaults to `false`), enable patching the Node's CJS loader when facing ESM errors, like `ERR_REQUIRE_ESM` in `node > v12.12.0`. 
+  - `useLibDir`: Prevents to compile lib folders on mocha runner
 
 ```json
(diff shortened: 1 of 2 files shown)


tomasmax (approved):

packages/sui-test/README.md
```diff
@@ -203,6 +203,7 @@ If defined, any error on your tests will create a screenshot of that moment in t
 - `server`: Config for `@s-ui/test server` binary:
   - `forceTranspilation`: List of regexs (string based, later will be transformed with `new Regex`) of modules to transpile. This is useful in case you're using server tests for modules that are ESModules based and need to be transpiled with `@babel/plugin-transform-modules-commonjs`.
   - `esmOverride`: Boolean flag (defaults to `false`), enable patching the Node's CJS loader when facing ESM errors, like `ERR_REQUIRE_ESM` in `node > v12.12.0`. 
+  - `useLibDir`: Prevents to compile lib folders on mocha runner
```
jcger, line 206: description might be a little bit confusing as it sounds like using this prop will prevent the lib folder from being compiled but as I'm understanding reading the code is something else. What about something like `activated by default. Prevents to compile lib folders on mocha runner if set to false`?

jcger (approved):

javiauso (approved):

kikoruiz (approved):

packages/sui-test/README.md
rmoralp: Agree with you. fixed

adriadoz (approved):

PR #1127, opened 2021-05-03
feat(packages/sui-theme): add margin xxs

PablitoGS:
## Description
add token xxs for margin

### DIFF
--- a/packages/sui-theme/src/settings/_spacing.scss
+++ b/packages/sui-theme/src/settings/_spacing.scss
@@ -18,6 +18,7 @@ $p-l: $p-base * 2 !default; // 16px
 $p-m: $p-base !default; // 8px
 $p-s: $p-base / 2 !default; // 4px
 $p-xs: $p-base / 4 !default; // 2px
+$p-xxs: $p-base / 8 !default; // 1px
 
 // Margins
 $m-base: 8px; // 8px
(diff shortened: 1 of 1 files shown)


packages/sui-theme/src/settings/_spacing.scss
```diff
@@ -30,6 +30,7 @@ $m-l: $m-base * 2 !default; // 16px
 $m-m: $m-base !default; // 8px
 $m-s: $m-base / 2 !default; // 4px
 $m-xs: $m-base / 4 !default; // 2px
+$m-xxs: 1px !default;
```
andresin87, line 33: IMO, all big sizes are relative to $m-base and it might also be as them.
`$m-xxs: $m-base / 8 !default; // 1px`

it might be also spreaded to paddings and add:
$p-xxs: $p-base / 8 !default; // 1px

turolopezsanabria (changes requested):
Can you please add the rationale for this request?

PablitoGS:
@turolopezsanabria ePreselec is using this design with an IJ component, it has a 1px gap between each list item

Captura de pantalla 2021-05-03 a las 13 03 40

PablitoGS:
@andresin87 , talking with @turolopezsanabria about this we think that it is a very particular case and it's not necessary to add this token. So I close this PR

turolopezsanabria:
Thnx!

PR #1130, opened 2021-05-04
fix(packages/sui-studio): style and context on query params persistance

andresin87:
## Description

`@s-ui/studio` uses queryParams for style and context navigation persistance through its different views
> PR targetting `@s-ui/studio-v9` branch


rmoralp:
Not sure about continue adding features to an older version. Could you pass also a PR to enables this feature last version of the studio?

andresin87:
> Not sure about continue adding features to an older version. Could you pass also a PR to enables this feature last version of the studio?

Sure, but i cannot test it linked

PR #1137, opened 2021-05-05
feat(packages/sui-critical-css): iterate critical css and remove cli

midudev:
Iterate Critical CSS:

- [x] Remove CLI for now to focus on programmatic usage.
- [x] Remove not needed deps.
- [x] Improve logging (for now, until we check this works everywhere).
- [x] Better exit on errors
- [x] Add middleware for @s-ui/ssr and Express


rmoralp (approved):

kikoruiz (approved):

packages/sui-critical-css/src/extract-from-app.js
```diff
@@ -27,23 +29,25 @@ export async function extractCSSFromApp({routes, config = {}}) {
 
     manifest[pathKey] = cssFileName
 
-    const configForMobileDevice = devices.m
-
     const css = await extractCSSFromUrl({
       url,
       ...configForMobileDevice
-    })
+    }).catch(() => '')
```
jcger, line 35: what about logging the error before returning an empty string? I'm afraid that we might hide an actual error

packages/sui-critical-css/src/middleware.cjs
```diff
@@ -0,0 +1,58 @@
+/* eslint-disable no-console */
+const fs = require('fs/promises')
+const path = require('path')
+const pathToRegexp = require('path-to-regexp') // peerDependency? ->
```
jcger, line 4: there is still that comment at the end that seems to be removable

jcger (approved):
we should plan adding tests sooner or later

PR #1151, opened 2021-05-14
Move @s-ui/svg to use vite instead @s-ui/bundler

midudev:
⚠️  While API is not changing, now icons are not transformed to React component but used as innerHTML instead. That means this **PR would be a Breaking Change.**

⚡  Use vite for @s-ui/svg instead @s-ui/bundler for almost instant dev mode.
🛑  Stop using @svgr and use directly svgo and avoid transforming SVG to React Component.
🌲  Use svg as innerHTML to improve React performance and avoid creating huge trees on VDOM.

🗜️  Reduction of dependencies of 50% (from 100MB to less than 50MB).
image

✅  Still working everything as expected (icons without colors are because they have fill attribute manually added).
image

Demo!

https://user-images.githubusercontent.com/1561955/118264165-9367cb80-b4b7-11eb-8c86-218267bdc974.mp4


desko27 (approved):
This is wonderful

oriolpuig (approved):

kikoruiz (approved):

adriadoz (approved):

packages/sui-svg/bin/sui-svg-build.js
```diff
@@ -46,15 +34,13 @@ const getLibFile = file => {
...
 const getAllSrcSvgFiles = () => fg([`${SVG_FOLDER}/**/*.svg`])
 
-const transformSvgToReactComponent = svg =>
-  svgr(svg, {
-    template,
-    expandProps: false,
-    removeTitle: true
-  })
+const transformSvgToReactComponent = svg => {
+  const {data} = optimize(svg)
```
rmoralp, line 38: This line only optimize the SVG.
After that, we execute template and create a React Component file which uses `dangerouslySetInnerHTML`. 
Given this generated component is a React Component but SVG has original attributes, like `xmlns:xlink` build could fail.

image

rmoralp (changes requested):

packages/sui-svg/bin/sui-svg-build.js
midudev: Awesome! Thanks for trying it on MA icons. :) I'll check it!

midudev: I think this problem is already fixed in latest alpha versions. :) Tried reinsalling it!

rmoralp (approved):

jcger (approved):

PR #1156, opened 2021-05-18
feat(packages/sui-critical-css): add new config to change filename

midudev:
Be able to generate different critical filenames. By default it would be `critical` but sometimes you might want to change this.


jcger (approved):

rmoralp (approved):

kikoruiz (approved):

kikoruiz:
@midudev You can close this PR because the problem you try to solve here is in fact already solved by the `outputDir` attribute (for multisites we also need to generate as many css files and manifests as sites we have defined).

PR #1157, opened 2021-05-18
feat(packages/sui-theme): add scss variable drop-down user height

ivanmlaborda:
## Description
Add default scss variable for avatar height in DropDown component. The actual avatar is rendering for the first time without a specific height generating annoying CLS.

## Related Issue
https://jira.scmspain.com/browse/MTR-37856


andresin87:
`settings-compat-v7` dir is no longer maintained and it will also be removed. This token might be placed in the right sui-component directory.

jordevo (approved):

adriadoz (approved):

AgonisticKatai (approved):

ivanmlaborda:
> `settings-compat-v7` dir is no longer maintained and it will also be removed (it may be). This token might be placed in the right sui-component directory.
DropdownUser is an adevinta-spain-component. Where is the place to put this kind of variables?

nucliweb (approved):

PR #1161, opened 2021-05-18
feat(packages/sui-ssr): Avoid using huge dependency for easy check

midudev:
Remove huge `parse-domain` dependency that's installing even Jest and some deprecated dependencies.

Also, this dependency fetch a ICANN list of correct domains to work.

We could avoid all of this with a simple check.


rmoralp (approved):

rmoralp:
Subdomain, local and forced to www tested in MA

PR #1181, opened 2021-05-31
fix(packages/sui-ssr): fix async stylesheet links replacement

kikoruiz:
## Description

Stylesheet links are not being replaced well in some of cases:
- If a stylesheet link is already async, it's being replaced again (duplicating some attributes)
- If HTML string has more than one link, it's only replaced the first one

This PR is fixing these 2 scenarios.


midudev (approved):

rmoralp (approved):

rmoralp:
Tested in MA

PR #1184, opened 2021-06-02
Feat/add display name support for critical css

rmoralp:
# Description
For some websites, it could be very hard to match a path-to-regex with all URL of a page.
We add support for using page displayName in middleware in order to simplify the page match.

Changes released in `@s-ui/critical-css@1.6.0-beta.2`

## Programatical usage:

In order to extract critical CSS and match extracted files with your page or route you can use two approaches:

- Using path-to-regex
- Using page component displayName

You can combine both of them, see readme updates

### DIFF
--- a/packages/sui-critical-css/README.md
+++ b/packages/sui-critical-css/README.md
@@ -15,7 +15,20 @@ Install package to your project:
 npm install @s-ui/critical-css -D
 ```
 
-Programmatic usage:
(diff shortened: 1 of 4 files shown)


rmoralp:
✅  Tested in MA

midudev (approved):

kikoruiz (approved):

PR #1208, opened 2021-06-22
fix(packages/sui-test): improve karma config construction

rmoralp:
## Description
Something happened with the previous release because it does not work as expected.
In this PR I have move it to the config file and tested it with beta `4.33.0-beta.1`

### DIFF
--- a/packages/sui-test/bin/karma/config.js
+++ b/packages/sui-test/bin/karma/config.js
@@ -2,10 +2,7 @@ const webpack = require('webpack')
 const path = require('path')
 const {clientConfig} = require('../../src/config')
 
-const captureConsole =
-  typeof clientConfig.captureConsole !== 'undefined'
-    ? clientConfig.captureConsole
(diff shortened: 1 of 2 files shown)


packages/sui-test/bin/karma/config.js
```diff
@@ -1,5 +1,11 @@
 const webpack = require('webpack')
 const path = require('path')
+const {clientConfig} = require('../../src/config')
```
midudev, line 3: Igual así se lee un poco mejor, no?
```
const {clientConfig} = require('../../src/config')
const {captureConsole = true} = clientConfig
```

midudev (approved):

packages/sui-test/bin/karma/config.js
rmoralp: Agree! Fixed

andresz1 (approved):

PR #1217, opened 2021-06-29
feat(packages/sui-ssr): allow to set a specific redirect status code

kikoruiz:
In some cases we should set a different kind of redirect status code: 302, 301 or even 307 if needed

### DIFF
--- a/packages/sui-ssr/server/ssr/index.js
+++ b/packages/sui-ssr/server/ssr/index.js
@@ -17,7 +17,6 @@ import ssrConfig from '../config'
 import {createStylesFor} from '../utils'
 import {getInitialContextValue} from '../initialContextValue'
 import {
-  statusCodes,
   redirectStatusCodes,
   DEFAULT_REDIRECT_STATUS_CODE
 } from '../../status-codes'
@@ -75,7 +74,7 @@ export default async (req, res, next) => {
       ? `?${qs.stringify(query)}`
       : ''
     const destination = `${redirectLocation.pathname}${queryString}`
-    return res.redirect(statusCodes.MOVED_PERMANENTLY, destination)
+    return res.redirect(DEFAULT_REDIRECT_STATUS_CODE, destination)
   }
 
   if (!renderProps) {
--- a/packages/sui-ssr/server/ssr/index.js
+++ b/packages/sui-ssr/server/ssr/index.js
@@ -16,6 +16,11 @@ import {buildDeviceFrom} from '../../build-device'
 import ssrConfig from '../config'
 import {createStylesFor} from '../utils'
 import {getInitialContextValue} from '../initialContextValue'
+import {
+  statusCodes,
+  redirectStatusCodes,
+  DEFAULT_REDIRECT_STATUS_CODE
+} from '../../status-codes'
 
 // __MAGIC IMPORTS__
 let contextProviders
@@ -28,7 +33,6 @@ try {
 // END __MAGIC IMPORTS__
 
 // const SERVER_TIMING_HEADER = 'Server-Timing'
-const HTTP_PERMANENT_REDIRECT = 301
 const HEAD_OPENING_TAG = '<head>'
 const HEAD_CLOSING_TAG = '</head>'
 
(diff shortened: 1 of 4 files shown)


packages/sui-ssr/status-codes.js
```diff
@@ -0,0 +1,11 @@
+export const statusCodes = {
+  MOVED_PERMANENTLY: 301,
+  FOUND: 302
+}
+
+export const DEFAULT_REDIRECT_STATUS_CODE = statusCodes.MOVED_PERMANENTLY
+
+export const redirectStatusCodes = [
```
kikoruiz, line 8: We only set here the codes we need to perform a redirect.

packages/sui-ssr/status-codes.js
```diff
@@ -0,0 +1,11 @@
+export const statusCodes = {
```
kikoruiz, line 1: This object can grow as we want in further iterations.

javiauso (approved):

rmoralp (approved):

jordevo (approved):
awesome! good job! 👏

andresz1 (approved):

packages/sui-ssr/status-codes.js
jordevo: `307` and `308` look like interesting options, too

`307`: _The Web page is temporarily unavailable for unforeseen reasons._
`308`: _Reorganization of a Web site, with non-GET links/operations._

https://developer.mozilla.org/en-US/docs/Web/HTTP/Redirections#temporary_redirections

packages/sui-ssr/server/ssr/index.js
```diff
@@ -165,9 +168,16 @@ export default async (req, res, next) => {
 
   const {__HTTP__} = initialProps
   if (__HTTP__) {
-    const {redirectTo} = __HTTP__
+    const {redirectTo, redirectStatusCode} = __HTTP__
```
andresz1, line 171: Maybe would be nice to have a generic `statusCode` in the future so `4XX` status codes could be also set

midudev (approved):

packages/sui-ssr/server/ssr/index.js
andresz1, line 171: I think that the `redirect` doesn't work in Client-Side. Any plan to add support in the future ?

andresz1:
IMO would be nice to add some documentation about how `redirectStatusCode` works and which will be the default value in case is not provided

packages/sui-ssr/server/ssr/index.js
andresz1: This is how [Next.js](https://nextjs.org/blog/next-10#notfound-support) does it

jordevo: I like the way Next.js does it, abstracting away the actual http status codes and setting them according to params (`notFound: true`, etc...)

kikoruiz: Let's see how this object grows, I think we can go ahead with this solution.

andresz1: Yup! would be nice since `getInitialProps` is intended to be isomorphic

andresz1: 👍🏻  in Milanuncios we need to use HTTP status error codes to tell crawlers that for example an ad is not found or has been deleted similar to what we do with the redirect!

```
Page.getInitialProps = () => {
  if (isNotFound) {
    return {
      __HTTP__: {
        notFound: true, 
       // or
       statusCode: 404 
      }
    }
  }
}
```

kikoruiz: @andresz1 But we're treating different things here, redirects are expected to work as it was defined before. You're suggesting to work with another kind of status codes, so that `res.redirect` maybe is not the ideal solution. Let's treat it in another topic, in fact, you have `earlyFlush` option in order to handle different response from requests made in the server.

kikoruiz:
@andresz1 Updated docs.

packages/sui-ssr/server/ssr/index.js
andresz1: Oh yes! sorry @kikoruiz this is not related with the `redirectStatusCode` feature! just something that would be nice to have in the future IMO

PR #1230, opened 2021-07-20
feat(packages/sui-pde): track feature flags and feature tests

jcger:
## Description
Tracking useFeature hook so we are able to analyse user behaviour based on a feature value. How will this work?
- on useFeature, we will send an Experiment Viewed with the feature key as experimentName and `On State` or `Off State` as variationName
- if a feature flag is used in a feature test, the tool will also send an Experiment Viewed for each linked experiment, working exactly the same way we track experiments with useExperiment

### DIFF
--- a/packages/sui-pde/src/hooks/useFeature.js
+++ b/packages/sui-pde/src/hooks/useFeature.js
@@ -2,6 +2,9 @@ import {useContext} from 'react'
 import PdeContext from '../contexts/PdeContext'
 import {getPlatformStrategy} from './platformStrategies'
 
+const VARIATION_NAME_ON = 'On State'
+const VARIATION_NAME_OFF = 'Off State'
+
 /**
  * track feature flag's own Experiment Viewed event
  * @param {object} param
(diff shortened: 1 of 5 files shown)


packages/sui-pde/src/hooks/useFeature.js
```diff
@@ -2,6 +2,45 @@ import {useContext} from 'react'
...
+ * @param {boolean} param.isActive
+ * @param {function} param.trackExperimentViewed
+ * @param {string} param.featureKey
+ */
+const trackFeatureFlagViewed = ({
+  isActive,
+  trackExperimentViewed,
+  featureKey
+}) => {
+  const variationName = isActive ? 'On State' : 'Off State'
```
kikoruiz, line 17: Maybe we could save those strings in two `const`

kikoruiz (approved):

midudev (approved):

rmoralp (approved):

rmoralp:
Great!

nucliweb (approved):

jordevo (approved):

kikoruiz:
"Experiment Viewed" tracks with needed properties are triggered properly, however forced variations are not working as expected regarding. So that, we'll merge this PR in order to be able to collect the tracks in our FF in PROD, but we'll keep investigating the issue with the forced variations.

PR #1238, opened 2021-08-13
feat(packages/sui-pde): force feature ssr support

andresz1:
## Description
Currently forcing a feature flag is not working with Server-Side Rendering. This PR will allow it.

How could it be done ?
```js
import {useLocation} from '@s-ui/react-router'
import {useFeature} from '@s-ui/pde'

const useEnhancedFeature = (key, attributes) => {
  const location = useLocation()
  return useFeature(key, attributes, location.search)
}

export default useEnhancedFeature
```
Since `useLocation` has Server-Side Rendering support this will work out of the box.

Another alternative would be to provide the location somehow when creating an instance of `PDE`

## Related Issue
None

### DIFF
--- a/packages/sui-pde/src/hooks/platformStrategies.js
+++ b/packages/sui-pde/src/hooks/platformStrategies.js
@@ -7,7 +7,14 @@ const getServerStrategy = () => ({
     return pde.getVariation({pde, name: experimentName, attributes})
   },
   trackExperiment: () => {},
-  getForcedValue: () => {}
+  getForcedValue: ({key, queryString}) => {
+    if (!queryString) {
+      return
+    }
+
+    const queryParams = parseQueryString(queryString)
+    return queryParams[`suipde_${key}`]
+  }
 })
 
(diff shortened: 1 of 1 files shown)


stivaliserna (approved):

nucliweb (approved):

javiauso (approved):

packages/sui-pde/src/hooks/platformStrategies.js
```diff
@@ -7,7 +7,14 @@ const getServerStrategy = () => ({
     return pde.getVariation({pde, name: experimentName, attributes})
   },
   trackExperiment: () => {},
-  getForcedValue: () => {}
+  getForcedValue: ({key, queryString}) => {
+    if (!queryString) {
```
andresz1, line 11: If there is no `queryString` we will do nothing just like before. No breaking change here!

midudev (approved):

adriadoz (approved):

arnau-rius (approved):

rmoralp (approved):

PR #1245, opened 2021-08-19
feat(packages/sui-i18n): add or modify translations dynamically

nistadev:
## Description
Added a new method to add translations dynamically and refresh the translator.

## Related Issue
N/A

## Example
It is useful if you are adding new translations dynamically. You can load less translations by default and then, when you navigate to a certain page, load the specific translations for that.


midudev (approved):

adriadoz (approved):
👏

andresin87 (approved):
👏

PR #1260, opened 2021-10-05
fix(packages/sui-react-initial-props): Be sure we have initial props when hydrating

midudev:
Fix wrong behaviour not returning initial props.

### DIFF
--- a/packages/sui-react-initial-props/src/withInitialProps.tsx
+++ b/packages/sui-react-initial-props/src/withInitialProps.tsx
@@ -3,18 +3,16 @@ import SUIContext from '@s-ui/react-context'
 import { RouteInfo } from '@s-ui/react-router/src/types'
 import { ClientPageComponent, WithInitialPropsComponent } from './types'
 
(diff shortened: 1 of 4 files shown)


packages/sui-react-initial-props/src/withInitialProps.tsx
```diff
@@ -52,7 +52,7 @@ export default (Page: ClientPageComponent): WithInitialPropsComponent => {
     const context = { ...suiContext, pathName: routeInfo.location.pathname }
 
     const [{ initialProps, isLoading }, setState] = useState(() => ({
-      initialProps: (initialPropsFromWindowRef.current != null) || {},
```
midudev, line 55: I think this was the line causing the problem.

kikoruiz (approved):

nucliweb (approved):

jcger (approved):

quinwacca (approved):

andresin87 (approved):

PR #1298, opened 2021-11-18
Fix prepare script to copy styles in generate template

paulusrex:
## Description
It fixes the npm script ```prepare``` in the template of a new generated component.

## Example
When ```npm run generate context component``` the script ```prepare``` in ```package.json``` will copy styles to ```lib``` folder.

### DIFF
--- a/packages/sui-studio/bin/sui-studio-generate.js
+++ b/packages/sui-studio/bin/sui-studio-generate.js
@@ -152,9 +152,9 @@ const defaultContext = `module.exports = {
 }
 `
 
-const prepare = swc
+const buildJs = swc
   ? 'sui-js-compiler'
-  : 'npm run build:js && npm run build:styles'
+  : 'babel --presets sui ./src --out-dir ./lib'
 
 // Check if the component already exist before continuing
 if (fs.existsSync(COMPONENT_PATH)) {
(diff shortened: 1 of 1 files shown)


andresz1 (approved):

packages/sui-studio/bin/sui-studio-generate.js
```diff
@@ -154,7 +154,7 @@ const defaultContext = `module.exports = {
 
 const prepare = swc
   ? 'sui-js-compiler'
-  : 'babel --presets sui ./src --out-dir ./lib'
+  : 'npm run build:js && npm run build:styles'
```
rmoralp, line 157: I think that the problem is different, we should only change js compilation if swc option is true
It means:
```
const buildJs = swc
  ? 'sui-js-compiler'
  : 'babel --presets sui ./src --out-dir ./lib'
```
And then:
```
"scripts": {
    "prepare": "npm run build:js && npm run build:styles",
    "build:js": "${buildJs}",
    "build:styles": "cpx './src/**/*.scss' ./lib"
  },
```

Isn't it? @midudev

paulusrex: Changed!

rmoralp (approved):

PR #1311, opened 2021-12-07
feat(packages/sui-pde): add decision listener to track segment events

rmoralp:


packages/sui-pde/src/adapters/optimizely/index.js
```diff
@@ -46,7 +49,12 @@ export default class OptimizelyAdapter {
     this._userId = userId?.toString()
     this._activeIntegrations = activeIntegrations
     this._applicationAttributes = applicationAttributes
+
     this.updateConsents({hasUserConsents})
+    this.addNotificationListener({
```
rmoralp, line 54: IMHO It could be moved and added on each website after adapter instance creation. It could be different tracking implementations (Segment, G.Analytics, etc)

PR #1352, opened 2022-02-10
Sui mono/workspace release config

jcger:
## Description
There might be repos with restricted and public packages. This repo allows us to use sui-mono config within each package to use its individual access info. In case the current package has no config, it will use the root one, if this one isn't set either it will use `restricted` by default as it did before.

## Example

Check out docs update

### DIFF
--- a/packages/sui-mono/test/server/configSpec.js
+++ b/packages/sui-mono/test/server/configSpec.js
@@ -70,7 +70,7 @@ describe('config', () => {
       expect(getPublishAccess()).to.equal('restricted')
     })
 
-    it('returns the access of the project if private', () => {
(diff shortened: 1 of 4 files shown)


packages/sui-mono/test/server/configSpec.js
```diff
@@ -70,7 +70,7 @@ describe('config', () => {
       expect(getPublishAccess()).to.equal('restricted')
     })
 
-    it('returns the access of the project if private', () => {
```
jcger, line 73: override this tests as duplicated

midudev (approved):

tonyghiani (approved):

PR #1383, opened 2022-03-03
Feat/add experimental loader esm

midudev:
Try to make sense with Sui-test server side and client side by adding file extensions but still transforming files.


jordevo (approved):

rmoralp:
What do you think about use a .nvmrc file to define node version? As we talked, those changes could mean that we will use a fixed node version here, isn't it?

packages/sui-test/src/descriptor-environment-patcher.js
```diff
@@ -27,58 +27,76 @@ export const descriptorsByEnvironmentPatcher = function descriptorsByEnvironment
...
+        throw new Error(
+          `Seems that you are doing a ${descriptorName}.${env}.only but you are running the tests for the ${
+            isNode ? environments.SERVER : environments.CLIENT
+          }\n\n`
+        )
+      }
+    } else {
+      return title =>
+        console.warn(
+          `skiping on the ${
```
jordevo, line 47:
```suggestion
          `skipping on the ${
```

sorry 🤓

jordevo (approved):

PR #1385, opened 2022-03-09
fix(packages/sui-mono): fix release for monopackage

rmoralp:
## Description
Fix release error when repo is monopackage (with no workspaces)
`✖ Error: Error: Root/package.json: ENOENT: no such file or directory, open 'Root/package.json'`

### DIFF
--- a/packages/sui-mono/src/config.js
+++ b/packages/sui-mono/src/config.js
@@ -6,7 +6,8 @@ const CWD = process.cwd()
 const packageFile = getPackageJson(CWD)
 
 const getWorkspaces = workspaces => {
-  if (workspaces.length === 0) return ['Root']
+  // If it is a monopackage, we return the current directory
+  if (workspaces.length === 0) return ['.']
   // If we have more than one workspace, we join
   // folders with the pattern {components/**,demo/**,tests/**}/package.json
   const pattern =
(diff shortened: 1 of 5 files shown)


packages/sui-mono/src/config.js
```diff
@@ -6,7 +6,8 @@ const CWD = process.cwd()
 const packageFile = getPackageJson(CWD)
 
 const getWorkspaces = workspaces => {
-  if (workspaces.length === 0) return ['Root']
```
jcger, line 9: isn't this gonna break the commit scopes if there as be no change in a workspace?

rmoralp: I think that `getWorkspaces` function should always the same, only return a list (array) of valid paths.
I have checked and fixed all side effects.

rmoralp: You could check all `getWorkspaces` usages here:
 https://github.com/SUI-Components/sui/search?q=getWorkspaces&type=code

jcger (approved):

rmoralp:
Beta published: `@s-ui/mono@2.22.0-beta.0`

kikoruiz (approved):

jordevo (approved):

PR #1389, opened 2022-03-25
feat(packages/sui-react-initial-props): redirect support

andresz1:
## Description
This PR:
- Avoid rendering page in Server-Side if `redirectTo` is defined. Besides saving time, this change avoids type errors since normally when performing a redirect no props are defined. See example:
```js
Page.getInitialProps = () => {
  if (!name) {
    return {
      __HTTP__: { // This will work in Client-Side and Server-Side
        redirectTo: '/path'
      }
    }
  }
  
  return {name}
}
```
- Add redirect support in Client-Side. Initially we are using `window.location` but we could also use the router in the future if needed.

We will add tests in the next pair programming session. Made with:
- @rmoralp 
- @jmanrumartinez 

## Beta
You can try it out installing the following beta package
- `@s-ui/react-initial-props@2.21.0-beta.0`

## Related Issue
None

## Example
See example above

### DIFF
--- a/packages/sui-react-initial-props/src/ssrComponentWithInitialProps.tsx
+++ b/packages/sui-react-initial-props/src/ssrComponentWithInitialProps.tsx
@@ -1,6 +1,6 @@
 import { renderToNodeStream, renderToString } from 'react-dom/server'
 import InitialPropsContext from './initialPropsContext'
-import {SsrComponentWithInitialPropsParams, InitialProps} from './types'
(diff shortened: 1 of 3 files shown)


midudev (approved):

jmanrumartinez (approved):

rmoralp (approved):

tonyghiani (approved):
Cool, this comes in handy on redirecting to the login page when the session has expired, good job. 👏🏻

PR #1394, opened 2022-03-30
feat(packages/sui-domain): Add a way to globally intercept errors on all axios instances

oegea:
This `PR` introduces a way to intercept all http responses and errors ocurred while using any `fetcher ` from the `sui-domain` package, by defining one single function that will be called every time an http error occurs.

### Why?

Recently in COCHES NET PRO, we needed a way to handle all http errors ocurred in all requests, to handle 401 status codes and redirect the user back to the login page when their token expired or became invalid.

Our first approach was to intercept errors by creating a facade class inside our domain package, that wrapped the sui-domain fetcher. 

This class, which we called `FetcherInterceptor`, exposed the same methods than the sui fetcher, and intercepted all errors ocurred during any http request, notifying them to a function in charge of error handling.

However, although this solution worked for us, our colleague @tomcask suggested us to implement something similar directly on the sui fetcher, in order to increase reusability.

### Why not using Axios Interceptors?

We wanted a solution decoupled from the final fetcher library, so if on a future we decide to use other third-party libraries that doesn't include interceptor features, we'll be sure that it doesn't affect to exceptions interception capabilities, and it won't require to implement interception logic every time a new fetcher is implemented.

### This suggestion

This PR suggests a way to wrap the standard `httpFetcher` inside a new class named `FetcherInterceptor`. This class, exposes the same methods than an standard fetcher, and ads two more: One to define the an error handle (the function that will be called if an error occurs), and another one to unset the error handler.

With these changes, two fetchers are exported: The standard `httpFetcher` and a new `interceptableHttpFetcher`, which is the one encapsulated by the `FetcherInterceptor` class.

## Related Issue

#[MTR-51774](https://jira.scmspain.com/browse/MTR-51774)

## Example

**How this works in COCHES NET PRO?**

We use an structure similar to this:

image

We have a single point in our web app in where we are able to intercept generic http errors. Mainly used nowadays to handle 401 errors.

This is an example of how a repository defines which function will intercept errors:

```
  interceptErrors({interceptHttpExceptionsRequest}) {
    const callback = interceptHttpExceptionsRequest.getCallback()
    this._fetcher._httpFetcher.setErrorInterceptor({callback})
  }
```

## Special Thanks ❤️

A lot of people contributed by planning this implementation and by giving ideas and feedback... (Apologies if I forget someone)

@tomcask 
@kikoruiz 
@davidmartin84 
@jordevo

### DIFF
--- a/packages/sui-domain/src/fetcher/AxiosFetcher.js
+++ b/packages/sui-domain/src/fetcher/AxiosFetcher.js
@@ -9,57 +9,6 @@ export default class AxiosFetcher {
    */
   constructor({config}) {
     this._axios = axios.create(config)
-    this._initInterceptors()
-  }
-
-  /**
-   * Initializes interceptors
-   */
-  _initInterceptors() {
-    this._axios.interceptors.response.use(
-      this._interceptResponse,
-      this._interceptError
-    )
-  }
-
-  /**
-   * Intercepts all responses and, in case an interceptor has been defined, calls it
-   * @param {Object} response
-   * @returns {Object}
-   */
-  _interceptResponse(response) {
-    if (AxiosFetcher.responseInterceptor !== undefined)
-      AxiosFetcher.responseInterceptor(response)
-
-    return response
-  }
-
-  /**
-   * Intercepts all errors and, in case an interceptor has been defined, calls it
-   * @param {Error} error
-   * @returns {Error}
-   */
-  _interceptError(error) {
-    if (AxiosFetcher.errorInterceptor !== undefined)
-      AxiosFetcher.errorInterceptor(error)
-
-    return error
-  }
-
-  /**
-   * Defines which function will be called when an error occurs
-   * @param {Function} interceptor
-   */
-  setErrorInterceptor(interceptor) {
-    AxiosFetcher.errorInterceptor = interceptor
-  }
-
-  /**
-   * Defines which function will be called when a response is received
-   * @param {Function} interceptor
-   */
-  setResponseInterceptor(interceptor) {
-    AxiosFetcher.responseInterceptor = interceptor
   }
 
   /**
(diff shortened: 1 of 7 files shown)


packages/sui-domain/src/fetcher/AxiosFetcher.js
```diff
@@ -9,6 +9,57 @@ export default class AxiosFetcher {
...
+    )
+  }
+
+  /**
+   * Intercepts all responses and, in case an interceptor has been defined, calls it
+   * @param {Object} response
+   * @returns {Object}
+   */
+  _interceptResponse(response) {
+    if (AxiosFetcher.responseInterceptor !== undefined)
```
jordevo, line 31: should you be checking for `typeof AxiosFetcher.responseInterceptor === 'function'`?

packages/sui-domain/src/fetcher/AxiosFetcher.js
```diff
@@ -9,6 +9,57 @@ export default class AxiosFetcher {
...
+    return response
+  }
+
+  /**
+   * Intercepts all errors and, in case an interceptor has been defined, calls it
+   * @param {Error} error
+   * @returns {Error}
+   */
+  _interceptError(error) {
+    if (AxiosFetcher.errorInterceptor !== undefined)
```
jordevo, line 43: same as above

jordevo (approved):

packages/sui-domain/src/fetcher/AxiosFetcher.js
oegea: I think we could add that to improve stability. However, `responseInterceptor` should only contain functions, defined by calling the `setResponseInterceptor` method.

However, if this `PR` is finally accepted, I'll add the extra type-checking before-merging.

Thank you!

kikoruiz (approved):

rmoralp (approved):

rmoralp:
@oegea is it possible to add some documentation in readme file?

AgonisticKatai (approved):

oegea:
> @oegea is it possible to add some documentation in readme file?

@rmoralp  Sure! I'll add some docs before merging. Thank you!

PR #1406, opened 2022-04-07
Feat/critical css add required classnames and retries

tomasmax:
## Description

Add `requiredClassNames` logic and retrier to critical css extractor:

- `requiredClassNames`: A list of required css class names. If they aren't present in the generated Critical CSS, it would be discarded. By default there would be 2 retries to try to get the correct Critical CSS
- `retries`: Number of retries if the requiredClassNames aren't present in the Cfritical CSS. By default it's 2.


javiauso (approved):

packages/sui-critical-css/src/extract-from-app.js
```diff
@@ -37,6 +37,41 @@ const waitForHealthCheck = async ({healthCheckUrl}) => {
...
+
+  if (!requiredClassNames) {
+    return css
+  }
+
+  const hasRequiredClasses = requiredClassNames.every(className =>
+    css?.includes(className)
+  )
+  if (!hasRequiredClasses) {
+    extratCriticalCSSFromUrlRetrier(
```
rmoralp, line 64: It needs a return to work as expected, isn't it?

packages/sui-critical-css/src/extract-from-app.js
```diff
@@ -65,12 +100,9 @@ export async function extractCSSFromApp({routes, config = {}}) {
 
     manifest[pathKey] = cssFileName
 
-    const css = await extractCSSFromUrl({
+    const css = await extratCriticalCSSFromUrlRetrier(pathOptions, {
```
rmoralp, line 103: IMHO it could have a generic name like `extractCSS` because of `extratCriticalCSSFromUrlRetrier` is coupled to implementation.

Also, what do you think about pass all the parameters as a named parameter?
```js
extratCriticalCSSFromUrlRetrier({url, configForMobileDevice, requiredClassNames, retries})
```

packages/sui-critical-css/src/extract-from-app.js
```diff
@@ -37,6 +37,42 @@ const waitForHealthCheck = async ({healthCheckUrl}) => {
   })
 }
 
+const extratCriticalCSS = async ({
```
jmanrumartinez, line 40:
```suggestion
const extractCriticalCSS = async ({
```

packages/sui-critical-css/src/extract-from-app.js
```diff
@@ -65,12 +101,12 @@ export async function extractCSSFromApp({routes, config = {}}) {
 
     manifest[pathKey] = cssFileName
 
-    const css = await extractCSSFromUrl({
+    const {requiredClassNames, retries} = pathOptions
+    const css = await extratCriticalCSS({
```
jmanrumartinez, line 105:
```suggestion
    const css = await extractCriticalCSS({
```

jmanrumartinez (approved):

rmoralp (approved):

PR #1415, opened 2022-05-05
fix(packages/sui-studio-create): Add missing comma in the JSON creator

crFront:
## Description
While using the sui-studio creator and installing it's dependencies we get an error since the generated JSON format is not correct. We're just missing a comma 😅

## Related Issue
None

## Example
Screenshot 2022-04-28 at 18 39 16


jordevo (approved):

jordevo:
good catch! 👌

ivanmlaborda (approved):

rmoralp (approved):

PR #1431, opened 2022-06-13
chore(packages/sui-test-e2e): Update Cypress to version 10.2.0

oegea:
## Description

👉 How to test in local: **Install the version `1.11.0-beta.1` of `@s-ui/test-e2e` package to test these changes.**
👉 How to test in CI: **Modify the CI job to use the `oegeaadevinta/e2e-test:cypress-10` docker image instead of `suitools/e2e-test`**

All of these should be temporary changes, just to check that your e2e tests run as expected after the update.

--

As part of the `end2end reborn` initiative, from the motor vertical, we have concluded that upgrading Cypress to the latest version available could be a good way to improve stability and performance of our end to end tests (among other actions), and additionally a way to get more useful features that aren't available on version `8.7.0`.

This PR introduces some changes to the `@s-ui/test-e2e` package to adopt the last Cypress version, which is the `10.2.0` version.

This is still a proof of concept, and we're currently performing some tests to ensure that tests keep working as expected after this changes.

[If you want to know more about this action and how it is related with the end2end reborn initiative, please refer to this document](https://docs.google.com/document/d/1odUPS0PCF3LlI60pAhBimN28vbuZnaYqJtJ97SA_iVk/edit) 

**Related to compatibility:**

Apparently, there are not breaking changes (we have tested it in coches.net and coches.net pro without modifying anything). However, as a side effect, a basic `cypress.config.js` file will be created on every project updated to Cypress 10. Until now we had the `configFile` parameter set to `false` to avoid creating configuration files, however, since Cypress 10, having that file seems to be mandatory. 

**Known Problems:**

* ✅  **SOLVED IN CYPRESS 10.2.0** Certificates validation errors: https://github.com/cypress-io/cypress/discussions/22128 - Seems that running tests on CI using chrome instead of Electron is a workaround. A fix has been implemented in Cypress and it's being merged and deployed, more info here: https://github.com/cypress-io/cypress/pull/22342

## Related Issue

* [#MTR-54607](https://jira.scmspain.com/browse/MTR-54607)


oegea:
> how about publishing a beta version of `sui-test-e2e` so we can try out these changes on local & CI?

Released the version `1.11.0-beta.0` for testing purposes. =D

oegea:
In order to test these changes in CI, I've tried to publish a docker image running the beta version, pushed without the `latest` tag to avoid collisions. However, I'm not able to push the image to Docker Hub due to a lack of credentials.

I'll keep posting updates here!

midudev (approved):

jmanrumartinez (approved):

jordevo (approved):

PR #1444, opened 2022-07-08
feat(packages/sui-bundler): Use new @parcel/css instead postcss and css minifier

midudev:
Faster and with less deps


jordevo (approved):
🚀

PR #1446, opened 2022-07-19
feat(packages/sui-ssr): Remove copy-webpack-plugin and rimraf as they're not needed

midudev:
Remove dependencies from sui-ssr to avoid installing them and use native alternatives.

### DIFF
--- a/packages/sui-ssr/bin/sui-ssr-build.js
+++ b/packages/sui-ssr/bin/sui-ssr-build.js
@@ -1,9 +1,9 @@
 #!/usr/bin/env node
 /* eslint no-console:0 */
 const program = require('commander')
-const rimraf = require('rimraf')
 const path = require('path')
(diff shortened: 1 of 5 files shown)


packages/sui-ssr/bin/sui-ssr-build.js
```diff
@@ -56,43 +58,53 @@ const build = () =>
...
+
+    webpack(nextConfig).run(async (error, stats) => {
+      if (error) return reject(error)
 
       const jsonStats = stats.toJson()
 
       if (stats.hasErrors()) {
-        return jsonStats.errors.map(error => console.log(error))
+        jsonStats.errors.map(error => console.log(error))
+        reject(new Error('Webpack build failed'))
```
arnau-rius, line 69: 👍🏽

PR #1450, opened 2022-08-08
feat(packages/sui-studio): Update template

andresin87:
## Description

Change the default template package-json component file npm run script


midudev:
👍 LGTM

PR #1460, opened 2022-09-01
feat(packages/sui-studio): Add new param to use explicit import

stivaliserna:
## Description
This PR adds a new option `--explicit` to generate a component in a file with its same name and an index re-export, this makes searching easier while keeping cleaner imports. 

```
Component
  src
    |--Component.js
    |--index.js
```

Beta: @s-ui/studio@11.21.0-beta.0

Feedback is much appreciated 🙂

### DIFF
--- a/packages/sui-studio/bin/sui-studio-generate.js
+++ b/packages/sui-studio/bin/sui-studio-generate.js
@@ -53,7 +53,7 @@ const componentInPascal = toPascalCase(
 const COMPONENT_DIR = `/components/${category}/${component}/`
 const COMPONENT_PATH = `${BASE_DIR}${COMPONENT_DIR}`
 const COMPONENT_ENTRY_JS_POINT_FILE = `${COMPONENT_PATH}src/index.js`
-const COMPONENT_JS_POINT_FILE = `${COMPONENT_PATH}src/${componentInPascal}.js`
+const EXPLICIT_COMPONENT_JS_POINT_FILE = `${COMPONENT_PATH}src/${componentInPascal}.js`
 const COMPONENT_PACKAGE_JSON_FILE = `${COMPONENT_PATH}package.json`
 const COMPONENT_PACKAGE_GITIGNORE_FILE = `${COMPONENT_PATH}.gitignore`
 const COMPONENT_PACKAGE_NPMIGNORE_FILE = `${COMPONENT_PATH}.npmignore`
@@ -249,7 +249,7 @@ test
     explicit ? explicitImportTemplate : componentTemplate
   ),
 
-  explicit && writeFile(COMPONENT_JS_POINT_FILE, componentTemplate),
+  explicit && writeFile(EXPLICIT_COMPONENT_JS_POINT_FILE, componentTemplate),
 
   writeFile(
     COMPONENT_ENTRY_SCSS_POINT_FILE,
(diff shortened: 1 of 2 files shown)


packages/sui-studio/bin/sui-studio-generate.js
```diff
@@ -52,6 +53,7 @@ const componentInPascal = toPascalCase(
 const COMPONENT_DIR = `/components/${category}/${component}/`
 const COMPONENT_PATH = `${BASE_DIR}${COMPONENT_DIR}`
 const COMPONENT_ENTRY_JS_POINT_FILE = `${COMPONENT_PATH}src/index.js`
+const COMPONENT_JS_POINT_FILE = `${COMPONENT_PATH}src/${componentInPascal}.js`
```
javiauso, line 56: Maybe a more descriptive name would help, something like `EXPLICIT_COMPONENT_JS_POINT_FILE`

javiauso (approved):

packages/sui-studio/bin/sui-studio-generate.js
stivaliserna: Agree! 👌 Thanks Javi!

packages/sui-studio/bin/sui-studio-generate.js
```diff
@@ -139,6 +141,25 @@ describe${context ? '.context.default' : ''}('${componentInPascal}', ${
...
+}
+
+${componentInPascal}.displayName = '${componentInPascal}'
+${componentInPascal}.propTypes = {}
+`
+
+const explicitImportTemplate = `import ${componentInPascal} from './${componentInPascal}.js'
+
+export default ${componentInPascal}
+`
```
arnau-rius, line 161: Maybe we can go with the shorter version that mixes the rexport of default and named exports.
```suggestion
const explicitImportTemplate = `export {default} from './${componentInPascal}.js'
`
```
Then if we want to export some named (for example: foo and bar) things inside the component would be much shorter:

`const explicitImportTemplate = `export {default, foo, bar} from './${componentInPascal}.js'`

arnau-rius:
If this is the new way of generating component, maybe we should create a script to modify also the already created components, so we just have one standard way of creating components. What do you think?

packages/sui-studio/bin/sui-studio-generate.js
stivaliserna: This is a proposal in stage 1, I checked and we use `@babel/plugin-proposal-export-default-from` to support it. If we're sure our infra will keep it, I'm ok with the change, but just want to double check 🙂

stivaliserna:
> If this is the new way of generating component, maybe we should create a script to modify also the already created components, so we just have one standard way of creating components. What do you think?

Yeah, makes sense! And we'll need to migrate the rest of components to actually solve our pain. Will work on it 👌 

Thanks @arnau-rius!

arnau-rius:
> > If this is the new way of generating component, maybe we should create a script to modify also the already created components, so we just have one standard way of creating components. What do you think?
> 
> Yeah, makes sense! And we'll need to migrate the rest of components to actually solve our pain. Will work on it 👌
> 
> Thanks @arnau-rius!

In case it gives you some guidance, here you can find two migrations that previously had been applied to the sui-studio, for cases similar to this one.

https://github.com/SUI-Components/sui/tree/master/packages/sui-studio/bin/migrations
- demo-folders
- test-folders

Maybe would be interesting to check that all marketplaces are OK with this new structure before coding/planning a migration.

👍🏽

PR #1465, opened 2022-09-28
fix(packages/sui-mono): add skip-ci as suffix

rmoralp:
## Description
Rollback this change https://github.com/SUI-Components/sui/commit/24763e652256613eef4d46460c206963def09e74
PR https://github.com/SUI-Components/sui/pulls?q=is%3Apr+is%3Aclosed

If we add `[skip ci]` as a prefix, we break commit naming convention and conventional-changelog does not work as expected with angular preset


oriolpuig:
❤️ 🙏🏻

Issue #1500, opened 2022-10-15
🪲 The hamburger menu icon is not visible on mobile


turolopezsanabria:
### Package

@sui-studio

### Description

The hamburger menu icon is not visible on mobile, although the button works as expected.

![221004 - The button is invisible on Mobile](https://user-images.githubusercontent.com/23620759/195999010-c527050a-bf05-4234-8335-b050837c1a44.png)

### Steps to Reproduce

1. Grab a phone (I've used an iPhone)
2. Go to [sui-components.vercel.app](https://sui-components.vercel.app/)
3. take a look at the menu

**Expected behavior:**

The hamburger menu should be visible on all devices

### Additional Information

The button is always visible on Desktop, even when you open developer tools in Chrome. You need to test with a Mobile device.

PR #1507, opened 2022-10-15
fix(packages/sui-studio): Add a fill to the buttons icon

turolopezsanabria:
`🔍 Show`

**Issue fixed with this PR**
https://github.com/SUI-Components/sui/issues/1500

## Description

The hamburger menu icon is not visible on mobile, although the button works as expected.

221004 - The button is invisible on Mobile

## Additional Information

The button is always visible on Desktop, even when you open developer tools in Chrome. You need to test with a Mobile device.

## Result after this PR

IMG_iPhone

### DIFF
--- a/packages/sui-studio/src/components/icons/index.js
+++ b/packages/sui-studio/src/components/icons/index.js
@@ -20,7 +20,7 @@ export const iconCode = (
 export const iconMenu = (
   <svg viewBox="0 0 24 24">
     <path d="M0 0h24v24H0z" fill="none" />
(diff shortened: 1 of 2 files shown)


turolopezsanabria:
Not sure this is actually fixing the problem :S
Although the color is now correct, I will keep searching for the source of the problem

turolopezsanabria:
OK fixed, ready for QA
Looks like Safari doesn't like SVG and Flex combination

arnau-rius (approved):

andresin87 (approved):

Issue #1522, opened 2022-10-19
Add prettier config file


Shubhamsahu19:
### Package

[name-of-the-package]@[version-of-the-package]

### Description

[Description of the issue]

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behavior:** [What you expect to happen]

**Actual behavior:** [What actually happens]

**Reproduces how often:** (only for bugs) [What percentage of the time does it reproduce?]

### Additional Information

Any additional information, configuration or data that might be necessary to reproduce the issue or in case of enhancement, describe why it would be useful.

PR #1523, opened 2022-10-19
Added .prettierrc config file

Shubhamsahu19:
#1522


rmoralp (changes requested):
We use sui-lint for that

Issue #1468, opened 2022-09-29
[sui-i18n] Move external dependency to package


rmoralp:
### Package

@s-ui/i18n

### Description

This package has `node-polyglot@0.4.3` as a dependency.
This version size is about 2.3kB minified

Given latest node-polyglot version (2.4.2) size is about 35.5kB minified
And current version achieve all our requirements
Then we have decided to move `node-polyglot@0.4.3` code to  `@s-ui/i18n` package and maintain this code.

### Additional Information

This task is not only about moving code to our package, we want to **keep our code style** and also **cover it with tests**

theonly1me:
@rmoralp Can you please assign me this issue?

PR #1528, opened 2022-10-22
feat(packages/sui-i18n): import node-polyglot 0.43 to our package

jordevo:
## Description
Import `node-polyglot 0.43` code to our package.

Polyglot adapter is already being fully tested:
https://github.com/SUI-Components/sui/blob/master/packages/sui-i18n/test/polyglotSpec.js#L53 

## Related Issue
Implements #1468 .

### DIFF
--- a/packages/sui-i18n/src/adapters/polyglot.js
+++ b/packages/sui-i18n/src/adapters/polyglot.js
@@ -23,28 +23,28 @@ const trim = str => {
 }
 
 const pluralTypes = {
-  chinese: function (n) {
+  chinese: n => {
     return 0
   },
-  german: function (n) {
+  german: n => {
     return n !== 1 ? 1 : 0
   },
-  french: function (n) {
+  french: n => {
     return n > 1 ? 1 : 0
   },
-  russian: function (n) {
+  russian: n => {
     if (n % 10 === 1 && n % 100 !== 11) return 0
     return n % 10 >= 2 && n % 10 <= 4 && (n % 100 < 10 || n % 100 >= 20) ? 1 : 2
   },
-  czech: function (n) {
+  czech: n => {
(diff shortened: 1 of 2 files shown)


packages/sui-i18n/src/adapters/polyglot.js
```diff
@@ -1,4 +1,222 @@
-import Polyglot from 'node-polyglot'
+// BSD 2-Clause License
+
+// Copyright (c) 2012, Airbnb
+// All rights reserved.
```
jordevo, line 4: ℹ️  license notice kept here as per:

https://github.com/airbnb/polyglot.js/blob/master/LICENSE

> 1. Redistributions of source code must retain the above copyright notice, this
>    list of conditions and the following disclaimer.

oegea: 👉 **A -personal- suggestion regarding this:** 

Would it make sense to have a specific file to maintain all third-party license notices? Relating each notice to the specific dependency that requires it.

oegea (approved):

packages/sui-i18n/src/adapters/polyglot.js
jordevo: good point! not sure what's the best way to deal with this in our packages.

rmoralp: We could open an issue to discuss this point and avoid blocking this PR. What do you think?

rmoralp (approved):

packages/sui-i18n/src/adapters/polyglot.js
jordevo: totally agree! will do that then 👍

PR #1533, opened 2022-11-09
feat(packages/sui-react-initial-props): hoist static props for server…

andresz1:
## Description
This PR hoist non react static props from page to server page this way they can also be used in Server-Side. For example:
in RE we are experimenting to track the SSR performance of pages and we want to enable it using a property in a page which is not available if it's no copied.

There are no breaking changes.
Beta available:
```
npm install @s-ui/react-initial-props@2.21.0-beta.1
```

## Related Issue
None

### DIFF
--- a/packages/sui-react-initial-props/package.json
+++ b/packages/sui-react-initial-props/package.json
@@ -17,10 +17,13 @@
     "react": "16 || 17"
   },
   "devDependencies": {
+    "@s-ui/react-router": "1",
     "@testing-library/react": "11.1.0",
     "@types/node": "16",
     "react": "17",
-    "react-dom": "17",
-    "@s-ui/react-router": "1"
+    "react-dom": "17"
+  },
+  "dependencies": {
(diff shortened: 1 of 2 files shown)


arnau-rius:
Is this lob compatible with React@17? https://github.com/mridgway/hoist-non-react-statics#compatible-react-versions

rmoralp (approved):

andresz1:
> Is this lib compatible with React@17? https://github.com/mridgway/hoist-non-react-statics#compatible-react-versions

In fact it is and we are already using it in our [router](https://github.com/SUI-Components/sui/blob/92a7bd5d06700cf7acd8866823ec733fd9513c02/packages/sui-react-context/src/index.tsx#L23). Nevertheless, the official support is for react 16 and including this package would also include [react-is@16](https://github.com/facebook/prop-types/blob/main/package.json#L31) and looks like there is a another for react 17. Will try to find an alternative

PR #1536, opened 2022-11-11
feat(packages/sui-studio): fix styles

andresin87:
## Description

- fix the padding affecting to the workbench area
- fix styles affecting to the workbench area elements
- SearchInput is fixed on top and it is not included in the scrollbar

## Example
image


turolopezsanabria (approved):
great PR!!
Comment, I think we don't need the inner shadow and padding of the menu.

jordevo (approved):

andresin87:
> great PR!! Comment, I think we don't need the inner shadow and padding of the menu.

removed

PR #1545, opened 2022-11-21
feat(packages/sui-bundler): mispelling error

andresin87:
## Description

mispelling error


packages/sui-studio/src/components/navigation/_style.scss
```diff
@@ -1,6 +1,5 @@
 .sui-StudioNav {
   border-radius: 4px;
-  box-shadow: inset 4px 4px 12px #dddddd;
```
andresin87, line 3: image
This line is no longer in production. Dont't know why its in the current master VCS branch

PR #1565, opened 2022-12-15
Create CLI `sui-jest` package to work with Jest

giodelabarrera:
Create `sui-jest` to work with Jest

### Why
After writing this discussion [Evaluar mejoras del entorno de test unitarios e integración](https://github.com/SUI-Components/sui/discussions/1553)
One possible solution that could resolve current pains and things to keep in mind is to try a new testing framework, we've chosen `jest` because it has the following points to highlight:
- run test files in parallel
- `jsdom` as fake DOM instead of chrome headless for unit/integration test level
- doesn't depend of Webpack, `swc` could be used as transformer
- provides a riched context when a test is failed
- create a sandbox for each test file
- provides features out-of-the-box like (mock, clock, exception, promises, ...)
- is standard and popular

### How
Creating a new `sui-jest` package. In that way, we could start to do more tries within a project (FotocasaPro) and later for others to retrieve feedback and more iterations.

The iteration adds the following:

- Creation of `sui-jest` package with:
  - Flexibility to use your own jest configuration. It will use a `jest.config.js` config file if this file exists within the project
  - Its required dependencies
- Add new glob test folder to ignore `**/__test__` in `sui-js-compiler`
- Add config for eslint jest env and jest plugin with `eslint-plugin-jest`

**Dependencies**

- `@swc/jest`
Transform test code using `swc` instead of `babel-jest`  for performance improvements
[@swc/jest – SWC](https://swc.rs/docs/usage/jest)

- `@types/jest`
Add jest typing allowing IDEs to recognize jest API
Screenshot 2022-12-14 at 08 57 24

- `jest-environment-jsdom`
Allow use `jsdom` as test environment of jest
[Configuring Jest · Jest](https://jestjs.io/docs/configuration#testenvironment-string)

- `read-pkg-up`
Allow to find the `package.json` of the repository consumer
[GitHub - sindresorhus/read-pkg-up: Read the closest package.json file](https://github.com/sindresorhus/read-pkg-up)

## Related Issue
[Evaluar mejoras del entorno de test unitarios e integración](https://github.com/SUI-Components/sui/discussions/1553)

### CLI API
`sui-jest [Jest CLI Options]`

### FotocasaPro demo

Having the script `test:jest` in our package.json 

Screenshot 2023-06-07 at 12 31 30

And the `jest.config.js` config file in our root folder

Screenshot 2022-12-15 at 14 03 19

We get the following result 🙂 

Screen Recording 2022-12-15 at 13 54 35


kikoruiz (approved):
We can move on like this, but I think we should put all this code into an independent package `sui-jest` or `sui-test-jest`, because including the Jest runner here doesn't give us too much benefit, it doesn't abstract us from the Jest specific implementation, moreover it adds more complexity to this current tool. @giodelabarrera WDYT?

giodelabarrera:
OK, in that case. In this pull request, I will move the `jest` code to a new package `sui-jest`. I prefer `sui-jest` than `sui-test-jest` for its simplicity.

Thanks in advance

packages/sui-js-compiler/package.json
```diff
@@ -19,4 +19,4 @@
     "fs-extra": "10.1.0",
     "typescript": "5.0.4"
   }
-}
+}
```
kikoruiz, line 22: I'd try to revert it for avoiding file conflicts...

kikoruiz (approved):
Good job! 👏 Maybe we could add a link into this table: https://github.com/SUI-Components/sui#available-packages

PR #1590, opened 2023-03-02
feat(packages/sui-i18n): Intl.FormatNumber extended options

AgonisticKatai:
## Description
In **coches.net**, we need the thousands points in the prices to be visible so as not to create confusion in the users of the price they are seeing. Currently the RAE standard says that it should only be in the numbers of more than 5 digits, but we believe that it is necessary to improve the reading of the data that is also in the thousands.

in order to meet this UX requirement, we added the possibility of being able to pass to **`Intl.NumberFormat`** a specific configuration to customize the type of formatting of the numbers according to the configuration described in **https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat/NumberFormat#options.**

## Related Issue
[MTR-66511](https://jira.ets.mpi-internal.com/browse/MTR-66511) & [MTR-64776](https://jira.ets.mpi-internal.com/browse/MTR-64776)

## Example
```javascript
import I18n from '@s-ui/i18n'
import Polyglot from '@s-ui/i18n/lib/adapters/polyglot'

const i18n = new I18n({adapter: new Polyglot()})

i18n.culture = 'es-ES'

// With default es-ES options
i18n.n(1000) //=> 1000
i18n.n(10000) //=> 10.000

// With forced useGrouping option
const defaultOptions = {
  useGrouping: true // should display grouping separators even if the locale prefers otherwise
}

i18n.defaultNumberFormatOptions = {...defaultOptions} 
i18n.n(1000) //=> 1.000
i18n.n(10000) //=> 10.000
```

### DIFF
--- a/packages/sui-i18n/README.md
+++ b/packages/sui-i18n/README.md
@@ -328,7 +328,7 @@ const defaultOptions = {
   useGrouping: true // should display grouping separators even if the locale prefers otherwise
 }
 
-i18n.defaultNumberFormatOptions = {...defaultOptions} 
+i18n.defaultNumberFormatOptions = defaultOptions
 i18n.n(1000) //=> 1.000
 i18n.n(10000) //=> 10.000
 ```
--- a/packages/sui-i18n/README.md
+++ b/packages/sui-i18n/README.md
@@ -324,11 +324,11 @@ i18n.n(1000) //=> 1000
 i18n.n(10000) //=> 10.000
 
 // With forced useGrouping option
-const extendedOptions = {
+const defaultOptions = {
   useGrouping: true // should display grouping separators even if the locale prefers otherwise
 }
 
-i18n.extendedIntlNumberOptions = {...extendedOptions} 
+i18n.defaultNumberFormatOptions = {...defaultOptions} 
 i18n.n(1000) //=> 1.000
 i18n.n(10000) //=> 10.000
 ```
--- a/packages/sui-i18n/README.md
+++ b/packages/sui-i18n/README.md
@@ -310,7 +310,7 @@ i18n.c(1000) //=> £1,000
 i18n.c(1000000) //=> £1,000,000
 ```
 
-We can format numbers with a extended Intl.FormatNumber options
+We can format numbers with a extended Intl.NumberFormat options
 ```javascript
 import I18n from '@s-ui/i18n'
 import Polyglot from '@s-ui/i18n/lib/adapters/polyglot'
--- a/packages/sui-i18n/README.md
+++ b/packages/sui-i18n/README.md
@@ -310,6 +310,29 @@ i18n.c(1000) //=> £1,000
 i18n.c(1000000) //=> £1,000,000
 ```
 
+We can format numbers with a extended Intl.FormatNumber options
+```javascript
+import I18n from '@s-ui/i18n'
+import Polyglot from '@s-ui/i18n/lib/adapters/polyglot'
+
+const i18n = new I18n({adapter: new Polyglot()})
+
+i18n.culture = 'es-ES'
+
+// With default es-ES options
+i18n.n(1000) //=> 1000
+i18n.n(10000) //=> 10.000
+
+// With forced useGrouping option
+const extendedOptions = {
+  useGrouping: true // should display grouping separators even if the locale prefers otherwise
+}
+
+i18n.extendedIntlNumberOptions = {...extendedOptions} 
+i18n.n(1000) //=> 1.000
+i18n.n(10000) //=> 10.000
+```
+
 #### Currency symbol
 
 Get the currency symbol.
--- a/packages/sui-i18n/README.md
+++ b/packages/sui-i18n/README.md
@@ -310,7 +310,7 @@ i18n.c(1000) //=> £1,000
 i18n.c(1000000) //=> £1,000,000
 ```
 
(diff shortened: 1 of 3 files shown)


oriolpuig (approved):
👏🏻

xmurcia (approved):

IgnacioND (approved):

jordevo (approved):

packages/sui-i18n/README.md
```diff
@@ -310,6 +310,29 @@ i18n.c(1000) //=> £1,000
 i18n.c(1000000) //=> £1,000,000
 ```
 
+We can format numbers with a extended Intl.NumberFormat options
```
jordevo, line 313:
```suggestion
We can format numbers with an extended Intl.NumberFormat options
```

packages/sui-i18n/src/i18n.js
```diff
@@ -7,6 +7,7 @@ export default class Rosetta {
     this._culture = null
     this._currency = null
     this._languages = null
+    this._extendedIntlNumberOptions = null
```
jordevo, line 10: how about using some more simple naming closer to what IMHO it actually does?

```suggestion
    this._defaultNumberFormatOptions = null
```

what do you think?

packages/sui-i18n/src/i18n.js
```diff
@@ -120,7 +127,10 @@ export default class Rosetta {
     }
 
     return typeof Intl !== 'undefined'
-      ? new Intl.NumberFormat(culture || this._culture, options).format(number)
+      ? new Intl.NumberFormat(culture || this._culture, {
+          ...options,
+          ...this._extendedIntlNumberOptions
```
jordevo, line 132: this way you would always be overriding `options` for each particular `.n` call with the preset `extendedIntlNumberOptions`...
 
perhaps I'm wrong but... how about this?

```suggestion
          ...this._extendedIntlNumberOptions,
          ...options
```

packages/sui-i18n/src/i18n.js
AgonisticKatai: Sorry, but i don't agree, IMHO this is values are an extended options which are added to those already received and set in the different methods that use i18n.n, for example i18n.c

```js
  c(number, minimumFractionDigits = 0, culture) {
    return this.n(
      number,
      {
        style: 'currency',
        currency: this._currency,
        minimumFractionDigits
      },
      culture
    )
  }
```

packages/sui-i18n/src/i18n.js
AgonisticKatai: I understand that this is precisely what is being sought with this change, to be able to override the current ones if necessary.

packages/sui-i18n/README.md
AgonisticKatai: done

packages/sui-i18n/src/i18n.js
jordevo: what I mean is...

image

in this case, the option `{useGrouping: false}` wouldn't do anything. Is it ok to override what's being explicitly declared when calling the `.n` method with the _more generic_ `extendedIntlNumberOptions`?

IMHO the more specific option should always have a way to override the _more generic_ one.

quinwacca: I agree with Jordi.

AgonisticKatai: ok, you've convinced me 😝

packages/sui-i18n/test/i18nSpec.js
```diff
@@ -156,6 +156,21 @@ describe('I18N', () => {
...
+        })
+        afterEach(() => {
+          i18n.culture = ''
+          i18n.extendedIntlNumberOptions = null
+        })
+
+        it('should display grouping separators even if the locale prefers otherwise', () => {
+          expect(i18n.n(1000)).to.eql('1.000')
+        })
+      })
```
jordevo, line 172: since you agreed on this https://github.com/SUI-Components/sui/pull/1590/files#r1122916373 

you could add another test to make sure options specified as args on `.n` do override the _generic_ ones 😉

packages/sui-i18n/README.md
```diff
@@ -310,6 +310,29 @@ i18n.c(1000) //=> £1,000
...
+// With default es-ES options
+i18n.n(1000) //=> 1000
+i18n.n(10000) //=> 10.000
+
+// With forced useGrouping option
+const defaultOptions = {
+  useGrouping: true // should display grouping separators even if the locale prefers otherwise
+}
+
+i18n.defaultNumberFormatOptions = {...defaultOptions}
```
arnau-rius, line 331:
```suggestion
i18n.defaultNumberFormatOptions = defaultOptions
```

arnau-rius (approved):

arnau-rius (commented):
Maybe it is interesting to add some tests to check that:
- Initial behaviour is conserved when options and defaultNumberFormatOptions are null.
- defaultNumberFormatOptions is `{useGrouping: true}` and options is`{useGrouping: false}` the defaultNumberFormatOptions takes preference over options

packages/sui-i18n/README.md
AgonisticKatai: done

PR #1598, opened 2023-04-27
feat(packages/sui-react-head): Wrap the Meta component to invert the …

marcbenito:
## Description
We've identified an issue with the preview of WhatsApp ads in one particular situation: sharing from an Android device.

When sharing from an iOS device, the preview appears as follows:
image

However, when sharing from an Android device, the preview is displayed like this:
image

We investigated and found the problem. Android doesn't recognize the meta property attribute unless it's the first attribute listed. This behavior occurs exclusively in server-side rendering (SSR).

Example:
image

If we move the data-rh attribute after property it works properly

image

This problem occurs on the following websites:

- Fotocasa
- Milanuncios
- Coches.net
- Motos.net

Currently, a version has been uploaded to Fotocasa (using a tag from the sui-react-head module) that solves the issue. Applying this fix would resolve the problem.

### Explanation of the technical solution:

We are using the react-head library, which exposes the Meta component:

image

Internally, this component injects the provided tag with the data-rh attribute at the beginning:

image

The proposed solution involves wrapping the tag component to reverse the order of the attributes.

### DIFF
--- a/packages/sui-react-head/src/index.tsx
+++ b/packages/sui-react-head/src/index.tsx
@@ -1,7 +1,9 @@
 /* eslint-disable react/prop-types */
 
 import * as React from 'react'
-import { HeadProvider, Link, Meta, Style, Title } from 'react-head'
+import {
+  HeadProvider, Link, Meta as MetaPrimitive, Style, Title
+} from 'react-head'
 
 import Body from './Body'
 import Html from './Html'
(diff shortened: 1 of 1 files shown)


packages/sui-react-head/src/index.tsx
```diff
@@ -23,6 +25,21 @@ interface HeadProps {
...
+  'data-rh': string
+}
+
+const MetaTagInverter: React.FC<MetaTagInverterProps> = ({ 'data-rh': rh, ...others }) => {
+  return <meta {...others} data-rh={rh} />
+}
+
+const Meta: React.FC<MetaProps> = (props) => {
+  // @ts-expect-error
+  return <MetaPrimitive {...props} tag={MetaTagInverter} />
```
marcbenito, line 40: we have to add the // @ts in order to solve typescript typing errors related to a "bad" implementation of the types definition in  react-head index.d.ts

andresz1 (approved):

kikoruiz (approved):

javiauso (approved):

PR #1603, opened 2023-05-17
feat(packages/sui-react-web-vitals): add routeid to reporter 

jordevo:
Add _routeid_ tag to Web Vitals Reporter as read from route `id` prop.

## Description
In order to provide an easier way to read Web Performance dashboards, we can now provide an `id` to a `route` that will be sent as `routeid` along with the other tags.

## Related Issue

N/A

## Example
```
        <Route path={HOME} id="home" getComponent={loadHomePage({appName})} />
```

### DIFF
--- a/packages/sui-react-web-vitals/src/index.js
+++ b/packages/sui-react-web-vitals/src/index.js
@@ -48,6 +48,12 @@ export default function WebVitalsReporter({
       return route?.path || route?.regexp?.toString()
     }
 
+    const getRoutename = () => {
+      const {routes} = router
+      const route = routes[routes.length - 1]
+      return route?.id || null
+    }
+
     const getDeviceType = () => {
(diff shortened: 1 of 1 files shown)


nucliweb (approved):

andresz1 (approved):

packages/sui-react-web-vitals/src/index.js
```diff
@@ -91,6 +99,14 @@ export default function WebVitalsReporter({
             key: 'pathname',
             value: getNormalizedPathname(pathname)
           },
+          ...(routename
+            ? [
+                {
+                  key: 'routename',
```
andresz1, line 105: `routeid` ? 🤔

jordevo: yeah perhaps it's more accurate

PR #1604, opened 2023-05-18
fix(packages/sui-lint): Fix Unknown rule error when checking scss files

ferransimon:
## Description

We are using an incompatible version between eslint and stylelint-config-recommended-scss causing the following error:
image

### DIFF
--- a/packages/sui-lint/package.json
+++ b/packages/sui-lint/package.json
@@ -38,7 +38,7 @@
     "prettier": "2.7.1",
     "stylelint": "14.11.0",
     "stylelint-config-prettier": "9.0.3",
-    "stylelint-config-recommended-scss": "9.0.1",
+    "stylelint-config-recommended-scss": "8.0.0",
     "stylelint-prettier": "2.0.0"
   }
 }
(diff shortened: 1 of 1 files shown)


jordevo (approved):

packages/sui-lint/package.json
```diff
@@ -38,7 +38,7 @@
     "prettier": "2.7.1",
     "stylelint": "14.11.0",
     "stylelint-config-prettier": "9.0.3",
-    "stylelint-config-recommended-scss": "7.0.0",
+    "stylelint-config-recommended-scss": "9.0.1",
```
jordevo, line 41: anything relevant on the changelog from major 7 to 8 and from major 8 to 9?

ferransimon: Checking it seems like the correct one to use is the 8 let me check if it works
image

ferransimon: It works, so I will downgrade it to 8 since we have the styleling-config at version 9 witch is the recommended one for the version 8

andresz1 (approved):

PR #1610, opened 2023-06-12
fix(packages/sui-i18n): fix missed key warning when i18n.url

javiauso:
## Description
Fix missed key warning when i18n.url.

image


jordevo:
fix? not sure this is a bug or a feature... 😅

javiauso:
> fix? not sure this is a bug or a feature... 😅

For now, it is warning with every `/` used in i18n.url because of converting it to empty string and checking if exists in the dictionary, and ir should't be like that.

jordevo (approved):

PR #1611, opened 2023-06-13
fix(packages/sui-i18n): restore slash when token is validated in i18n.url

javiauso:
## Description
Restore first slash when token is validated in i18n.url.

### DIFF
--- a/packages/sui-i18n/test/i18nSpec.js
+++ b/packages/sui-i18n/test/i18nSpec.js
@@ -79,6 +79,12 @@ describe('I18N', () => {
         expect(i18n.n(1000000)).to.eql('1,000,000')
       })
 
+      it('generate url properly', () => {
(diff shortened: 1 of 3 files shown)


packages/sui-i18n/test/i18nSpec.js
```diff
@@ -79,6 +79,12 @@ describe('I18N', () => {
         expect(i18n.n(1000000)).to.eql('1,000,000')
       })
 
+      it('generate url properly', () => {
+        expect(i18n.url('/profilePathSegment/privacyPathSegment')).to.eql(
+          '/my-profile/privacy'
+        )
+      })
```
jordevo, line 86: 👏

jordevo (approved):

lolarufino (approved):

carlosvillu (approved):

PR #1618, opened 2023-08-03
feat(packages/sui-react-web-vitals): improve LCP and INP reporting

jordevo:
## Description
Adding necessary features so that `reportAllChanges: true` can be configured for specific metrics, which provides more fine-grained data about **CWV** metrics. By default, setting it for `LCP` and `INP`. Also, by importing the `web-vitals/attribution` build, we're also able to collect the elements of the page related to the recorded performance metrics.

Enabling custom thresholds for metrics via props to allow for fine-tuning of performance metrics, reduce noise on reporting or simply go with Google's defined thresholds for **CWV**.

## Related Issue

n/a

### DIFF
--- a/packages/sui-react-web-vitals/src/index.js
+++ b/packages/sui-react-web-vitals/src/index.js
@@ -8,19 +8,23 @@ import useMount from '@s-ui/react-hooks/lib/useMount/index.js'
 import {useRouter} from '@s-ui/react-router'
 
 export const METRICS = {
-  TTFB: 'TTFB',
-  LCP: 'LCP',
   CLS: 'CLS',
+  FCP: 'FCP',
   FID: 'FID',
   INP: 'INP',
-  FCP: 'FCP'
+  LCP: 'LCP',
+  TTFB: 'TTFB'
 }
 
 const DEFAULT_METRICS_REPORTING_ALL_CHANGES = [METRICS.LCP, METRICS.INP]
 
 const DEFAULT_CWV_THRESHOLDS = {
+  [METRICS.CLS]: 100,
+  [METRICS.FCP]: 1800,
+  [METRICS.FID]: 100,
+  [METRICS.INP]: 200,
   [METRICS.LCP]: 2500,
-  [METRICS.INP]: 200
+  [METRICS.TTFB]: 800
 }
 
 export const DEVICE_TYPES = {
--- a/packages/sui-react-web-vitals/src/index.js
+++ b/packages/sui-react-web-vitals/src/index.js
@@ -60,7 +60,7 @@ export default function WebVitalsReporter({
       return deviceType || browser?.deviceType
     }
 
-    const handleChange = ({name, value}) => {
+    const handleChange = ({attribution, name, value}) => {
       const onReport = onReportRef.current
       const pathname = getPathname()
       const routeid = getRouteid()
--- a/packages/sui-react-web-vitals/src/index.js
+++ b/packages/sui-react-web-vitals/src/index.js
@@ -16,7 +16,12 @@ export const METRICS = {
   FCP: 'FCP'
 }
 
-const METRICS_REPORTING_ALL_CHANGES = [METRICS.LCP, METRICS.INP]
+const DEFAULT_METRICS_REPORTING_ALL_CHANGES = [METRICS.LCP, METRICS.INP]
+
+const DEFAULT_CWV_THRESHOLDS = {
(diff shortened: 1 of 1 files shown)


packages/sui-react-web-vitals/src/index.js
```diff
@@ -58,7 +60,60 @@ export default function WebVitalsReporter({
...
+      const isExcluded =
+        !pathname || (Array.isArray(pathnames) && !pathnames.includes(pathname))
+
+      if (isExcluded) {
+        return
+      }
+
+      const amount = name === METRICS.CLS ? value * 1000 : value
+
+      logger.log({
```
nucliweb, line 76: ## What do you think about defining a threshold to avoid sending good INP values to 🍄?

### Define the threshold following the Core Web Vitals values

```js
const CWV_THRESHOLDS = {
  LCP: 2500,
  INP: 200
}
```

```suggestion
      const isOverThreshold = value > CWV_THRESHOLDS[name]
      isOverThreshold && logger.log({
```

jordevo: good point, indeed. Will refactor accordingly ASAP!

packages/sui-react-web-vitals/src/index.js
```diff
@@ -58,7 +60,60 @@ export default function WebVitalsReporter({
...
+        name: 'cwv',
```
andresz1, line 77: The metric should be have a different name since `cwv` is already distribution

packages/sui-react-web-vitals/src/index.js
```diff
@@ -58,7 +60,60 @@ export default function WebVitalsReporter({
...
+        amount,
+        tags: [
```
andresz1, line 79: Tags are not needed

packages/sui-react-web-vitals/src/index.js
andresz1: Perhaps thresholds configuration could be a prop ? WDYT ?

jordevo: good point too. We could have Google's thresholds as default values and add a `prop` so they can be overridden... IMHO it can be useful for further perf fine-tuning.

hpintos (approved):
👏🏼👏🏼👏🏼

packages/sui-react-web-vitals/src/index.js
```diff
@@ -129,13 +160,23 @@ export default function WebVitalsReporter({
 
 WebVitalsReporter.propTypes = {
   /**
-   * An optional array of core web vitals. Choose between: TTFB, LCP, FID, CLS and INP. Defaults to all.
+   * An object with METRICS as keys and thresholds as values
+   * Thresholds by default are those above which Google considers the page as "needs improvement"
+   * Lower thresholds could be set for fine-tuning, higher thresholds could be set for less noise when reporting all changes
    */
-  metrics: PropTypes.arrayOf(PropTypes.oneOf(Object.values(METRICS))),
+  cwvThresholds: PropTypes.object,
```
andresz1, line 167: What about just `thresholds`

oriolpuig (approved):
❤️

packages/sui-react-web-vitals/src/index.js
jordevo: better naming, indeed! done ✅

packages/sui-react-web-vitals/src/index.js
```diff
@@ -58,17 +71,31 @@ export default function WebVitalsReporter({
...
+      const amount = name === METRICS.CLS ? value * 1000 : value
+      const pathname = getPathname()
+      const isExcluded =
+        !pathname || (Array.isArray(pathnames) && !pathnames.includes(pathname))
+
+      if (isExcluded || !logger?.log || amount < thresholds[name]) return
+
+      logger.log(
+        JSON.stringify({
+          name: `cwv|${name.toLowerCase()}`,
```
andresz1, line 84: `cwv.${name.toLowerCase()}` 🙏🏻 let's always use dot notation for naming log/metric events

packages/sui-react-web-vitals/src/index.js
```diff
@@ -58,17 +71,31 @@ export default function WebVitalsReporter({
...
+          tags: {amount, ...attribution}
```
andresz1, line 85: `tags` key value is not needed you can merge it directly in the object above

andresz1 (approved):

PR #1620, opened 2023-08-11
feat(packages/sui-bundler): add fast refresh support

andresz1:
## Description
This PR add fast refresh support in development only (for Client-Side rendering)
- This will improve the DX

This should be added to the `app.js`

```js
if (module.hot) {
  module.hot.accept()
}
```


kikoruiz (approved):
I guess we'll need to adapt these changes when we have merged all the TS support changes.

andresz1:
> I guess we'll need to adapt these changes when we have merged all the TS support changes.

Absolutely! there is way to make it work with swc

carlosvillu (approved):

oriolpuig (approved):

jordevo (approved):
👏

PR #1622, opened 2023-08-14
chore(Root): stop sending devhose metrics

kikoruiz:
## Description
Not needed since we're not reviewing/using those metrics at all.


jordevo (approved):
🚀 hack-celerate FTW

andresz1 (approved):

PR #1636, opened 2023-09-05
fix(packages/sui-react-web-vitals): use metricsAllChanges and add more default metrics

nucliweb:

### DIFF
--- a/packages/sui-react-web-vitals/src/index.js
+++ b/packages/sui-react-web-vitals/src/index.js
@@ -16,6 +16,13 @@ export const METRICS = {
   TTFB: 'TTFB'
 }
 
+// https://github.com/GoogleChrome/web-vitals#metric
+const RATING = {
+  good: 'good',
+  needsImprovement: 'needs-improvement',
+  poor: 'poor'
+}
+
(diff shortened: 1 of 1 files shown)


andresz1 (approved):

jordevo (approved):

carlosvillu (approved):

packages/sui-react-web-vitals/src/index.js
```diff
@@ -16,7 +16,19 @@ export const METRICS = {
...
+  needsImprovement: 'needs-improvement',
+  poor: 'poor'
+}
+
+const DEFAULT_METRICS_REPORTING_ALL_CHANGES = [
+  METRICS.CLS,
+  METRICS.FID,
+  METRICS.INP,
+  METRICS.LCP
+]
```
jordevo, line 31: I'd be a bit more careful with this and would leave just one or two metrics we as a whole really want to focus on to be reported on all changes by default. Perhaps I'd just leave `INP` there... what do you think? 

The idea precisely behind the `metricsAllChanges` prop is to allow the user of this reporter component to customize which CWV metrics to report on all changes.

PR #1637, opened 2023-09-05
feat(packages/sui-react-web-vitals): leave just INP to be reported on all changes

jordevo:
## Description
Leave it to every consumer of the component to configure which metrics will report _on all changes_.
Since metric values are already being reported when above specific thresholds, reporting _on all changes_ above and below said thresholds by default can be a bit too much data.

## Related Issue
https://github.com/SUI-Components/sui/pull/1636


nucliweb:
We want all the Core Web Vitals with values to improve, right?

jordevo:
> We want all the Core Web Vitals with values to improve, right?

yes, just not by default and leaving it to the consumer of the Reporter's choice. What do you think?

nucliweb (approved):

jordevo:
leaving all metrics by default to report on all changes because after the latest changes on the reporter, the data is being collected only when over a certain threshold, which is optimal enough

Issue #1601, opened 2023-05-04
[sui-pde] avoid tracking Experiment Viewed on feature flags


alextremp:
### Package

pde

### Description

Currently when we request for a feature flag to be active or not, if the user was not assigned to an optimizely response, we always track to Segment the `Experiment Viewed` event.

We have a request to avoid that as much as we can to avoid consuming `Segment API Calls` when there's no analytic interest on the feature flags.

As a proposal, and following a previous open issue to use structured args in useFeature ( #1418 ), we should add an additional arg to the useFeature:

- `track`: Boolean, `false` by default. Setting it to true would mean there is a specific analytic interest on a feature flag and it should track the `Experiment Viewed` event. Letting it to false, it should not track the event to Segment.

Note that this would mean to push a Major version or send a clear notification about the change to current PDE users as the default tracking behavior will change.

**Expected behavior:** Only track the `Experiment Viewed` events to Segment if the new `track` parameter is explicitly set to `true` into the `useFeature` parameters

**Actual behavior:** `Experiment Viewed` events are always tracked to Segment

PR #1641, opened 2023-09-28
[feature][sui-pde] Remove Experiment Viewed event by default.

carlosvillu:
## Description
In order to reduce unnecessary calls to Segment, the Experiment Viewed event is disabled by default.

If you need to track how many times your experiment has been viewed, you should set the shouldTrackExperimentViewed argument to true.

## Related Issue
https://github.com/SUI-Components/sui/issues/1601

## Example

```js
    const {isActive} =  useFeature(
            'featureKey4',
            {attribute1: 'value'},
            undefined,
            undefined,
            true
          )
```

### DIFF
--- a/packages/sui-pde/README.md
+++ b/packages/sui-pde/README.md
@@ -229,6 +229,18 @@ const optimizelyAdapter = new OptimizelyAdapter({
 })
 ```
 
+#### Track Experiment Viewed
+
+In order to reduce unnecessary calls to Segment, the `Experiment Viewed` event is disabled by default.
+
+If you need to track how many times your experiment has been viewed, you should set the `shouldTrackExperimentViewed` argument to true.
+
+```js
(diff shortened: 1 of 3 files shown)


jordevo (approved):

jordevo:
not a breaking change per se but we should make everyone aware of this change...

kikoruiz (approved):

carlosvillu:
> not a breaking change per se but we should make everyone aware of this change...

I spoke with Carles Ferrer, and we agreed that next week he would bring this matter up in the cross-domain meeting.

Initially, it's more of a data-related issue than a frontend concern. However, it does make sense to post a message in a public channel so that everyone is aware.

The plan is to merge the PR next week to avoid having it in production over the weekend.

nucliweb (approved):

PR #1654, opened 2023-11-02
feat(packages/sui-react-router): avoid extra render during first load

andresz1:
## Description
This PR avoid an extra re-render in the `Router` component. Normally this causes the application to re-render twice during first load

## Related Issue
None

## Example
None

### DIFF
--- a/packages/sui-react-router/src/Router.js
+++ b/packages/sui-react-router/src/Router.js
@@ -65,7 +65,7 @@ const Router = ({
   const [state, setState] = useState({router, params, components})
 
   useEffect(() => {
-    let prevState = {}
+    let prevState = null
     const handleTransition = (err, nextState) => {
       if (err) {
         if (onError) return onError(err)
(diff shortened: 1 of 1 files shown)


jordevo (approved):

javiauso (approved):

AgonisticKatai (approved):

oriolpuig (approved):

nachoAdevinta (approved):

packages/sui-react-router/src/Router.js
```diff
@@ -66,15 +66,26 @@ const Router = ({
 
   useEffect(() => {
     let prevState = {}
+    let isSkipped = !!matchContext
```
andresz1, line 69: If `matchContext` is provided that means that we can avoid the `setState` being execute since the router properties are going to be the same

kikoruiz (approved):

nucliweb (approved):
👏🏼

PR #1651, opened 2023-10-20
feat(packages/sui-lint): update line length to 120 characters

jordevo:
## Description
New screens are wider and have larger resolutions so it seems it's about time to be able to read our code on longer than 80-character lines.

Code chunk with 80-character lines:
image

Same code chunk with 120-character lines:
image

## Related Issue
N/A

## Example
N/A

PR #1665, opened 2023-11-21
[REFACTOR] apply new linting rules

kikoruiz:
apply new linting rules added on https://github.com/SUI-Components/sui/pull/1651


jordevo (approved):

nucliweb (approved):

kikoruiz:
> could've been a `chore` type of commit in order to skip the triggering of all the packages' releases...

Not sure about this fact... the only thing I know is this one will be funnier 😆 

Anyway, I've created another one with a `chore` type of commits, so we can choose.

andresz1:
> could've been a `chore` type of commit in order to skip the triggering of all the packages' releases...

I agree with @jordevo

PR #1669, opened 2023-11-23
feat(packages/sui-react-web-vitals): add sub inp metrics to cwv

andresz1:
## Description
This PR:
- Add INP metric sub-division tracking (input delay, processing time and presentation delay) to have better visibility of INP issues
- Add tests
- Remove not needed code
- Now if there is a route id the pathname will not be tracked using the distribution method

Breaking changes:
- Changed `pathnames` prop with `allowed` (an optional array of pathnames or route ids that you want to track). This will allow us to filter by route ids.

This change will require a new major

Screenshot 2023-11-23 at 17 16 01

## Example

Screenshot 2023-11-23 at 17 12 46


jordevo (approved):
🔝 glad to see some new tests! 👏

packages/sui-react-web-vitals/src/index.js
```diff
@@ -91,51 +80,82 @@ export default function WebVitalsReporter({
...
 
       logger.cwv({
         name: `cwv.${name.toLowerCase()}`,
         amount,
-        path: hasPathOnRoute ? pathname : getNormalizedPathname(pathname),
+        path: pathname,
         target,
         loadState: attribution.loadState,
-        ...(attribution.eventType && {eventType: attribution.eventType}),
-        visibilityState: document.visibilityState
```
andresz1, line 111: This should be added back

packages/sui-react-web-vitals/src/index.js
```diff
@@ -91,51 +80,82 @@ export default function WebVitalsReporter({
...
       }
 
       if (!logger?.distribution) return
 
       const amount = name === METRICS.CLS ? value * 1000 : value
+      const tags = [
+        ...(routeid
+          ? [
+              {
+                key: 'routeid',
```
nucliweb, line 140: 👏

packages/sui-react-web-vitals/src/index.js
```diff
@@ -91,51 +80,82 @@ export default function WebVitalsReporter({
       }
     }
 
+    const computeINPMetrics = entry => {
```
nucliweb, line 83:
```suggestion
    // RenderTime is an estimate because duration is rounded and may get rounded down.
    // In rare cases, it can be less than processingEnd and that breaks performance.measure().
    // Let's ensure it's at least 4ms in those cases so you can barely see it.
    const computeINPMetrics = entry => {
```

nucliweb (approved):
🚀

nucliweb:
👀  Remember to update the package.json to upgrade to a major version

packages/sui-react-web-vitals/src/index.js
andresz1: 👍🏻  I just added it

carlosvillu (approved):

PR #1672, opened 2023-11-29
feat(packages/sui-test-contract): make jobs execution halt on contract publish errors

kikoruiz:
## Description
When there are some errors executing `sui-test-contract publish` command in CI, the job where the command is being executed will never fail neither stop. So we need to explicitly throw an error.


jordevo (approved):

oriolpuig (approved):
Interesting!

carlosvillu (approved):

nucliweb (approved):

andresz1 (approved):

PR #1673, opened 2023-11-30
feat(packages/sui-bundler): add server dev config

andresz1:
## Description
This PR adds an initial webpack configuration to enable Server-Side Rendering Fast Refresh support during development

### DIFF
--- a/packages/sui-bundler/webpack.config.server.dev.js
+++ b/packages/sui-bundler/webpack.config.server.dev.js
@@ -1,0 +1,119 @@
+// @ts-check
+
+const path = require('path')
+const webpack = require('webpack')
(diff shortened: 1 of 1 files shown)


javiauso (approved):

andresz1:
> I wonder if this should go...?
> 
> https://github.com/SUI-Components/sui/blob/593b9028706b58070a2614fbcd04d28131a0d890/packages/sui-bundler/webpack.config.server.js#L18

This file is specific to work with Server-Side rendering in development mode. We can consider a merge in the future but let's keep it isolated while we perform some tests

nucliweb (approved):

kikoruiz (approved):

PR #1702, opened 2024-01-12
fix(packages/sui-mono): fix package name regex

kikoruiz:
This PR solves the current issue we have in the master build: https://github.com/SUI-Components/sui/actions/runs/7500241868/job/20418527951#step:12:22

build error


packages/sui-mono/src/check.js
```diff
@@ -20,7 +20,7 @@ const DEPS_UPGRADE_COMMIT_TYPE = 'upgrade'
 const DEPS_UPGRADE_COMMIT_TYPE_TO_PUSH = 'feat'
 const DEPS_UPGRADE_PACKAGES = ['deps', 'deps-dev']
 const DEPS_UPGRADE_BRANCH_PREFIX = 'dependabot/npm_and_yarn/'
-const SCOPE_REGEX = /packages\/[a-z]+-[a-z]+/
```
kikoruiz, line 23: The old regex was only matching a string structure like `packages/sui-any`

jordevo (approved):

nucliweb (approved):

andresz1 (approved):

PR #1724, opened 2024-02-08
feat(packages/sui-react-web-vitals): add some device data to sent to Open Search

nucliweb:
## Description

Add `deviceMemory`, `effectiveType`, and `hardwareConcurrency` device information to send to Open Search.

PR #1725, opened 2024-02-08
feat(packages/sui-react-web-vitals): adding device perf params to logger.cwv tracks

ivanmlaborda:
Adding device perf params to logger.cwv tracks

## Description
To distinguish interactions based on connection quality, and device performance in OpenSearch CWV logs we need to add these parameters to the logged data.

**effectiveConnection, deviceMemory, hardwareConcurrency**

_example data sent to Mushroom_
imatge

This data can be calculated and setted in suiContextFactory in our webapps

```
const hardwareConcurrency = isClient && navigator.hardwareConcurrency
const effectiveConnection = isClient && navigator.connection?.effectiveType
const deviceMemory = isClient && navigator.deviceMemory
```
must be added to browser property in the Context
```
{
  ...
  browser: {
    ...
    effectiveConnection,
    deviceMemory,
    hardwareConcurrency
  }
}

```

### DIFF
--- a/packages/sui-react-web-vitals/src/index.js
+++ b/packages/sui-react-web-vitals/src/index.js
@@ -103,6 +103,8 @@ export default function WebVitalsReporter({
 
       const target = getTarget({name, attribution})
 
+      const {effectiveConnection, deviceMemory, hardwareConcurrency} = browser || {}
+
       logger.cwv({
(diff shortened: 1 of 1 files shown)


nucliweb:
With this #1724 you don't need change the web app.

PR #1734, opened 2024-03-05
feat(packages/sui-js): Add sha256 hash method

ivanmlaborda:
## Description
We are adding the sha256 hash method to sui-js utilities. This new function uses browser native methods so we are not coupling it to third-party libraries.
This method will be async, so we are naming it createAsyncSha256.

https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/digest#converting_a_digest_to_a_hex_string

## Example
import {createAsyncSha256} from '@s-ui/js/lib/hash'

const hashedValue = await createAsyncSha256(value)


carlosvillu (approved):

oriolpuig (approved):

jordevo (approved):

jordevo:
pretty safe to use with our main user base...

image

PR #1740, opened 2024-03-15
Add Hydration Mismatch Overlay for dev mode

jordevo:
Bringing *builder.io*'s [Hydration overlay](https://github.com/BuilderIO/hydration-overlay/) tool to help us debug hydration mismatch errors on dev environment.

## Description
Using `withHydrationOverlayWebpack` in web pack development configuration allows us to spot hydration mismatches and therefore prevent introducing potencial re-renderings on client and hence, performance issues.

image

This [additional change](https://github.com/SUI-Components/sui/pull/1740/files#diff-3c38d2c065de8c4205a1963bf98939999af9b4b6ff3826f15ce6cea72bb7c5ebR90) is necessary to make it work. The hydration mismatch overlay only shows up when it detects errors containing the words `hydration` or `hydrating` (see here https://github.com/BuilderIO/hydration-overlay/blob/661aa0e17a50777fe9d7222fe0759d84798b1c3e/packages/lib/src/hydration-overlay-initializer.js#L5 ). Since the `app.js` script is being served from `localhost:8080` whereas the dev server is running on `localhost:3000`, an error from `app.js` would be coming from a different origin and pop up as a generic `Script error`, which would not be captured by `hydration-overlay`. Hence the requirement for a new dependency `html-webpack-inject-attributes-plugin`.

### DIFF
--- a/packages/sui-bundler/webpack.config.client.dev.js
+++ b/packages/sui-bundler/webpack.config.client.dev.js
@@ -6,6 +6,7 @@ const HtmlWebpackPlugin = require('html-webpack-plugin')
 const {WebpackManifestPlugin} = require('webpack-manifest-plugin')
 const ReactRefreshWebpackPlugin = require('@pmmmwh/react-refresh-webpack-plugin')
 const MiniCssExtractPlugin = require('mini-css-extract-plugin')
+const HtmlWebpackInjectAttributesPlugin = require('html-webpack-inject-attributes-plugin')
 const {withHydrationOverlayWebpack} = require('@builder.io/react-hydration-overlay/webpack')
 
 const {envVars, MAIN_ENTRY_POINT, config, cleanList, when} = require('./shared/index.js')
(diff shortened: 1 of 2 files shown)


hpintos (approved):
🚀

packages/sui-bundler/webpack.config.client.dev.js
```diff
@@ -83,7 +85,10 @@ const webpackConfig = {
     }),
     new WebpackManifestPlugin({fileName: 'asset-manifest.json'}),
     new webpack.HotModuleReplacementPlugin(),
-    new ReactRefreshWebpackPlugin({overlay: false})
+    new ReactRefreshWebpackPlugin({overlay: false}),
+    new HtmlWebpackInjectAttributesPlugin({
+      crossorigin: 'anonymous'
```
jordevo, line 90: This is necessary to make it work. The hydration mismatch overlay only shows up when it detects errors containing the words hydration or hydrating (see here https://github.com/BuilderIO/hydration-overlay/blob/661aa0e17a50777fe9d7222fe0759d84798b1c3e/packages/lib/src/hydration-overlay-initializer.js#L5 ). Since the app.js script is being served from localhost:8080 whereas the dev server is running on localhost:3000, an error from app.js would be coming from a different origin and pop up as generic Script error, which would not be captured by hydration-overlay. Hence the requirement for a new dependency html-webpack-inject-attributes-plugin.

carlosvillu (approved):

marcbenito (approved):
❤️

kikoruiz (commented):
🙌

PR #1746, opened 2024-03-27
feat(packages/sui-decorators): Create AsyncInlineError decorator

oriolpuig:
## Description
Frontend Tech Community has agreed to create, use, and standardize the `@AsyncInlineError()` decorator. It will improve our code readability. 

This decorator catch-all method errors and each method instead of return a value, will return a tuple like `[error, response]`. So, **with this decorator a method using it will never fail**. 

On this PR, we provide: 
- A new `@AsyncInlineError()` decorator
- Update the needed documentation
- New linting rules created to push our ADR decisions. 
  - MUST The `@AsyncInlineError()` must be the first decorator in a method. It will make our life easier when we develop other decorators. The rule provide a quick fix.
  - SHOULD The `@AsyncInlineError()` should be used in all UseCase.execute, Service.execute and Repository.method. The rule will NOT provide a quick fix to avoid uncontrolled errors. 

```javascript 

@AsyncInlineError()
function execute() {
  if(error) {
    throw new Error('this is not ok')
  }
  return 'this is ok'
}

const response = execute()
const [error, result] = response
``` 

### Client test ✅
image

### Server test ✅
image


carlosvillu (approved):

packages/eslint-plugin-sui/src/rules/decorators-adr.js
```diff
@@ -0,0 +1,79 @@
...
+          )
+
+          // TODO validar en services
+          // TODO validar en repositories
+          // TODO quickFix implementar
+          if (!asyncInlineErrorDecoratorNode) {
+            context.report({
+              node: executeFn.key,
+              messageId: 'notFoundAsyncInlineErrorDecorator',
+              fix: fixer => {
```
oriolpuig, line 59: Remove code suggestion/fix to avoid uncontrolled code mutation.

kikoruiz (approved):

PR #1749, opened 2024-04-05
Fix/copy html files dev ssr

andresz1:
## Description
This PR copies `500.html` and `404.html` files

### DIFF
--- a/packages/sui-ssr/bin/sui-ssr-dev.js
+++ b/packages/sui-ssr/bin/sui-ssr-dev.js
@@ -3,6 +3,7 @@
 
 const program = require('commander')
 const {exec} = require('child_process')
+const {copyFile} = require('fs/promises')
 const path = require('path')
 const fs = require('fs')
 const express = require('express')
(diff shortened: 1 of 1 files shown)


jordevo (approved):

carlosvillu (approved):

packages/sui-ssr/bin/sui-ssr-dev.js
```diff
@@ -63,6 +66,13 @@ const linkStatics = () => {
   )
 }
 
+const copyStatics = () => {
+  return Promise.allSettled([
+    cp(path.join(SRC_PATH, '404.html'), path.join(PUBLIC_OUTPUT_PATH, '404.html'), {recursive: true}),
+    cp(path.join(SRC_PATH, '500.html'), path.join(PUBLIC_OUTPUT_PATH, '500.html'), {recursive: true})
```
jordevo, line 72: this should be ok if one of the files (or both) doesn't exist, right?

andresz1: yep because allSettled is used

nucliweb (approved):

PR #1753, opened 2024-04-23
Add more custom Lint Rules for the Golden Path

carlosvillu:
- [x] CommonJS syntax [More info](https://nodejs.org/api/modules.html#modules-commonjs-modules)
- [x] Domain decorators adoption (number of decorators)


andresz1 (approved):

packages/eslint-plugin-sui/src/rules/commonjs.js
```diff
@@ -0,0 +1,110 @@
...
+      url: 'https://github.mpi-internal.com/scmspain/es-td-agreements/blob/master/30-Frontend/00-agreements'
+    },
+    fixable: null,
+    schema: [],
+    messages: {
+      forbiddenExports: dedent`
+        Use module.* should be avoid.
+      `,
+      forbiddenRequires: dedent`
+        Use require function should be avoid.
```
kikoruiz, line 28: What if anyone is using a `require` in a ESM compliant file? e.g.:
```
import { createRequire } from "module"
const require = createRequire(import.meta.url)

require('whatever')
```

Should we allow it or not?

kikoruiz: Doc reference: https://nodejs.org/api/module.html#modulecreaterequirefilename

PR #1754, opened 2024-04-29
Feature/update webpack dev middleware versions

xmurcia:
## Description

Use patched versions of `webpack-dev-middleware` to avoid vulnerabilities.
https://github.com/webpack/webpack-dev-middleware/security/advisories/GHSA-wr3j-pwj9-hqq6

### DIFF
--- a/packages/sui-bundler/package.json
+++ b/packages/sui-bundler/package.json
@@ -51,7 +51,7 @@
     "strip-ansi": "6.0.1",
     "style-loader": "3.3.1",
     "url": "0.11.0",
-    "webpack-dev-server": "4.10.0",
+    "webpack-dev-server": "5.0.4",
     "webpack-manifest-plugin": "5.0.0",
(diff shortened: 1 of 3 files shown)


packages/sui-bundler/package.json
```diff
@@ -51,7 +51,7 @@
     "strip-ansi": "6.0.1",
     "style-loader": "3.3.1",
     "url": "0.11.0",
-    "webpack-dev-server": "4.10.0",
+    "webpack-dev-server": "5.0.4",
```
kikoruiz, line 54: @andresz1 You maybe know better if this change has important implications in our implementation, here are the breaking changes: https://github.com/webpack/webpack-dev-server/blob/master/migration-v5.md

xmurcia: I have just used the version 4.15.2 which is just a patched version fixing the vulnerability. This should work without any problems.

kikoruiz (approved):

PR #1756, opened 2024-05-03
Domain Fetcher Circuit Breaker & Circuit Breaker Manager

jordevo:


packages/sui-domain/src/circuitBreaker/CircuitBreaker.js
```diff
@@ -0,0 +1,65 @@
...
+  HALF: 'HALF'
+}
+
+export class CircuitBreaker {
+  request = null
+  state = CircuitBreakerStates.CLOSED
+  failureCount = 0
+  failureThreshold = 5 // number of failures to determine when to open the circuit
+  resetAfter = 50000
+  timeout = 5000 // declare request failure if the function takes more than 5 seconds
```
jordevo, line 13: these default settings can be overridden by passing an `options` object to the `CircuitBreakerManager` constructor

PR #1764, opened 2024-05-21
feat(packages/sui-domain): adapt entry point factory to be able to wo…

AgonisticKatai:
According to the new [domain agreements](https://github.mpi-internal.com/scmspain/es-td-agreements/blob/master/30-Frontend/00-agreements/01-domain.md#creational-pattern-agreements), use cases must be exported in a named way by means of a constant called `factory` that will be responsible for giving access to the `create method` of the use case.

## Description
This change must maintain backward compatibility with the current entry point behavior.

## Related Issue
N/A

## Example
```js
class GetMovieDetailsUseCase extends UseCase {
  /* ... */
  static create({config}) {
    return new GetMovieDetailsUseCase({repository: MovieRepository.create({config})})
  }
}

export const factory = GetMovieDetailsUseCase.create
```

### DIFF
--- a/packages/sui-domain/src/EntryPointFactory.js
+++ b/packages/sui-domain/src/EntryPointFactory.js
@@ -1,5 +1,14 @@
 import createNotImplementedUseCase from './createNotImplementedUseCase.js'
 
+const METHODS_BY_FACTORY_TYPES = {
+  // Return a method from a whole default exported factory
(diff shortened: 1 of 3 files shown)


emiliovillu (approved):
❤️

oriolpuig (approved):
💅🏻

jordevo (approved):

kikoruiz (approved):

PR #1767, opened 2024-06-06
feat(packages/sui-studio): Add jest compatibility with test path on studio-dev

jelowin:
## Description

At Fotocasa Portal, we are integrating Jest into our tests. One of the problems we encountered was that we couldn't start the demos with studio-dev because it was searching for the path "test/index.test", leading to an error.
Screenshot 2024-06-06 at 10 14 29

This PR solves the problem by detecting if the component has a Jest test and then changing the test alias to later import the correct path.

## Example

Demo on component with Jest test

Screenshot 2024-06-07 at 13 46 05

### DIFF
--- a/packages/sui-studio/bin/sui-studio-dev.js
+++ b/packages/sui-studio/bin/sui-studio-dev.js
@@ -45,8 +45,10 @@ if (!category || !component) {
 
 const componentPath = path.join(PWD, 'components', category, component)
 const legacyTestPath = path.join(PWD, 'test', category, component)
+const jestPath = path.join(componentPath, '__tests__')
 
 const testPath = fs.existsSync(legacyTestPath) ? legacyTestPath : path.join(componentPath, 'test')
+const isJestTest = fs.existsSync(jestPath)
 
 const {cache, ...others} = config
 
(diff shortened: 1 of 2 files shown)


packages/sui-studio/bin/sui-studio-dev.js
```diff
@@ -45,21 +45,26 @@ if (!category || !component) {
...
+    ...config.plugins,
+    new webpack.DefinePlugin({__COMPONENT_ID__: JSON.stringify(componentID), __DISABLE_TESTS__: isJestTest})
+  ],
   resolve: {
     ...config.resolve,
     alias: {
       ...config.resolve.alias,
       component: path.join(componentPath, 'src'),
-      test: testPath,
+      test: isJestTest ? jestPath : testPath,
```
andresz1, line 67: I think that now this is not needed

carlosvillu (approved):

kikoruiz (approved):

packages/sui-studio/bin/sui-studio-dev.js
oriolpuig: Totally agree.

oriolpuig (approved):

packages/sui-studio/bin/sui-studio-dev.js
jelowin: @andresz1 I only keep the testPath, isn't it?

`test: testPath`

PR #1775, opened 2024-06-19
feat(packages/sui-js): add method to check if user agent is a bot

ivanmlaborda:
Check user agents to detect bots

## Description
checkLegitimateCrawler method can detect if the passed user agent is one of the following legitimate bots
[  
  'googlebot',
  'google-structured-data-testing-tool',
  'bingbot',
  'linkedinbot',
  'mediapartners-google',
  'debugbear'
]

## Example
const BOT_USER_AGENT = 'Mozilla/5.0 (compatible; Googlebot/2.1; +http://www.google.com/bot.html)'

checkUserAgentIsBot(BOT_USER_AGENT) // returns true


packages/sui-js/src/bots/index.js
```diff
@@ -1,4 +1,4 @@
-const BOTS_USER_AGENTS = [
+const LEGITIMATE_BOTS_USER_AGENTS = [
```
jordevo, line 1:
```suggestion
const LEGITIMATE_BOT_USER_AGENTS = [
```

jordevo (approved):

jordevo:
we have a `ua-parser` context in this package... wouldn't it make sense to keep this `checkUserAgentIsBot` util there? 🤔 

https://github.com/SUI-Components/sui/blob/master/packages/sui-js/src/ua-parser/index.js

crFront (approved):

AgonisticKatai (approved):

kikoruiz (approved):

PR #1789, opened 2024-07-19
Feat/scmi 124866 private attribute lint rule

aleixadevinta:
## Description
In this pull request we ensure that a doamin model class (value object or entity) has all attributes as private and each attribute has his getter.

## Related Issue
[SCMI-124866](https://jira.ets.mpi-internal.com/browse/SCMI-124866)

### DIFF
--- a/packages/eslint-plugin-sui/src/rules/private-attributes-model.js
+++ b/packages/eslint-plugin-sui/src/rules/private-attributes-model.js
@@ -1,0 +1,90 @@
+/**
+ * @fileoverview ensure domain model have a private attributes and each attribute has a getter
+ */
+'use strict'
+
+const dedent = require('string-dedent')
+
+// ------------------------------------------------------------------------------
+// Rule Definition
+// ------------------------------------------------------------------------------
+
+/** @type {import('eslint').Rule.RuleModule} */
+module.exports = {
+  meta: {
+    type: 'suggestion',
+    docs: {
+      description: 'Ensure domain models have all attributes as private and each attribute has a getter',
+      recommended: false,
+      url: 'https://github.mpi-internal.com/scmspain/es-td-agreements/blob/master/30-Frontend/00-agreements'
+    },
+    fixable: null,
+    schema: [],
+    messages: {
+      attributeHasToBePrivate: dedent`
+        If your class is a domain model (Value Object or Entity), you have to define all attributes as private.
+      `,
+      privateAttributeHasToHaveGetter: dedent`
+        If your class is a domain model (Value Object or Entity), you have to define a getter for the attribute #{{attributeName}}.
+        You can define a native getter (get {{attributeName}}) or a custom getter ({{customGetterName}}).
+      `
+    }
+  },
+
+  create(context) {
+    return {
+      ClassDeclaration(node) {
+        const className = node?.id?.name ?? ''
+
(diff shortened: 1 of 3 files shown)


packages/eslint-plugin-sui/src/rules/private-attributes-model.js
```diff
@@ -0,0 +1,90 @@
...
+
+        privateAttributes.forEach(attribute => {
+          let hasGetter = false
+          const customGetterName = `get${attribute.key.name.charAt(0).toUpperCase()}${attribute.key.name.slice(1)}`
+
+          classMethods.forEach(method => {
+            const existNativeGetterWithAttributeKey = method.key.name === attribute.key.name && method.kind === 'get'
+            const existCustomGetterWithAttributeKey = method.key.name === customGetterName
+
+            if (existNativeGetterWithAttributeKey | existCustomGetterWithAttributeKey) {
```
oriolpuig, line 71: I'm not sure if the condition is well, it's intentionally with one `|` or should it be `||`?

```suggestion
            if (existNativeGetterWithAttributeKey || existCustomGetterWithAttributeKey) {
```

packages/eslint-plugin-sui/src/rules/private-attributes-model.js
```diff
@@ -0,0 +1,90 @@
...
+      `
+    }
+  },
+
+  create(context) {
+    return {
+      ClassDeclaration(node) {
+        const className = node?.id?.name ?? ''
+
+        const allowedWords = ['VO', 'ValueObject', 'Entity']
```
oriolpuig, line 39: This is a good approach and this rule will cover a lot of cases, but it will not check classes without names but under the correct place (`valueObjects` or `entities` folders). 

To achieve that, with the lines below, you can check if the file is inside those folders and run this rule inside them too. 

```javascript
const filePath = context.getFilename()
    const relativePath = path.relative(context.getCwd(), filePath)

    // Check if the file is inside the requierd folders (entities, services, repositories, ...)
    const entitiesPattern = /entities|Entities/i
    const isEntityPath = entitiesPattern.test(relativePath)

    // ... same with value objects
```

See [this PR lines](https://github.com/SUI-Components/sui/pull/1746/files#diff-bcd92c5b18a0f6057bbf74129b80f233d3eb3a5c97c1b3f9652dedf53b8c2e34R41-R52), and follow the code to understand. 

I suggest you detect different cases in our code base and run it there to see if this rule is working fine in different scenarios.

oriolpuig (approved):
Great job!

packages/eslint-plugin-sui/src/rules/private-attributes-model.js
```diff
@@ -0,0 +1,90 @@
...
+        publicAttributes.forEach(attribute => {
+          context.report({
+            node: attribute,
+            messageId: 'attributeHasToBePrivate'
+          })
+        })
+
+        // Check if a private attribute has a public accessor
+        const privateAttributes = node.body.body.filter(node => {
+          return node.type === 'PropertyDefinition' && node.key.type === 'PrivateIdentifier'
```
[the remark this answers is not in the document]
aleixadevinta: We decided in the agreement that all private attributes will be defined by #, as it is the native way to define private attributes in the latest versions of ECMAScript.

kikoruiz (approved):

PR #1837, opened 2024-10-09
feat(packages/sui-pde): add decide listener

andresz1:
## Description
This PR:
- Fix an issue when consents are not accepted returning an object
- Fix issue where `ExperimentViewed` event was being sent for feature flags only

### DIFF
--- a/packages/sui-pde/src/hooks/useDecision.js
+++ b/packages/sui-pde/src/hooks/useDecision.js
@@ -39,6 +39,16 @@ export default function useDecision(name, {attributes, trackExperimentViewed, qu
         return {enabled: true, flagKey: name, variationKey: forced}
       }
 
+      const notificationId = pde.addDecideListener({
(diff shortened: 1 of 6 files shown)


packages/sui-pde/src/hooks/useDecision.js
```diff
@@ -39,20 +39,24 @@ export default function useDecision(name, {attributes, trackExperimentViewed, qu
         return {enabled: true, flagKey: name, variationKey: forced}
       }
 
+      const notificationId = pde.addDecideListener({
+        onDecide: ({type, decisionInfo: decision}) => {
+          const {ruleKey, variationKey, decisionEventDispatched} = decision
+
+          if (type === 'flag' && decisionEventDispatched) {
```
andresz1, line 46: This will make sure that `experimentViewed` is only sent when the feature test has a linked experiment

PR #1852, opened 2024-10-21
feat(packages/sui-segment-wrapper): set ga user id on identify

kikoruiz:
Sets GA `user_id` on every `identify` call to Segment Analytics.

### DIFF
--- a/packages/sui-segment-wrapper/src/repositories/googleRepository.js
+++ b/packages/sui-segment-wrapper/src/repositories/googleRepository.js
@@ -53,12 +53,10 @@ const getGoogleField = async field => {
 export const getGoogleClientID = () => getGoogleField(FIELDS.clientId)
 export const getGoogleSessionID = () => getGoogleField(FIELDS.sessionId)
 
-export const sendGoogleUserId = userId => {
+export const setGoogleUserId = userId => {
   const googleAnalyticsMeasurementId = getConfig('googleAnalyticsMeasurementId')
 
   if (!googleAnalyticsMeasurementId || !userId) return
 
-  window.gtag?.('set', googleAnalyticsMeasurementId, {
-    user_id: userId
-  })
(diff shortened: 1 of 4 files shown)


andresz1 (approved):

packages/sui-segment-wrapper/src/repositories/googleRepository.js
```diff
@@ -52,3 +52,13 @@ const getGoogleField = async field => {
...
+
+export const sendGoogleUserId = userId => {
+  const googleAnalyticsMeasurementId = getConfig('googleAnalyticsMeasurementId')
+
+  if (!googleAnalyticsMeasurementId || !userId) return
+
+  window.gtag?.('set', googleAnalyticsMeasurementId, {
+    user_id: userId
+  })
+}
```
jordevo, line 64: looking at this I'd say it's just setting the `userId`, not sending it... am I right?

```suggestion
export const setGoogleUserId = userId => {
  const googleAnalyticsMeasurementId = getConfig('googleAnalyticsMeasurementId')

  if (!googleAnalyticsMeasurementId || !userId) return

  window.gtag?.('set', googleAnalyticsMeasurementId, {
    user_id: userId
  })
}
```

jordevo (approved):

AgonisticKatai (approved):

carlosvillu (approved):

nucliweb (approved):

packages/sui-segment-wrapper/src/repositories/googleRepository.js
kikoruiz: @jordevo Yep, in fact it's incorrect, I've changed the implementation (and also the function name) following the official docs: https://developers.google.com/analytics/devguides/collection/ga4/reference/config#user_id_ex

PR #1857, opened 2024-10-29
Feat decide toggle

andresz1:
## Description
This PR:
- Enables `useDecision` to prevent sending impression tracking events to both Optimizely and Segment by setting `isEventDisabled` to `true`
- Introduces a new hook, `useDecisionCallback`, which allows for programmatic calls to decide, making it useful for user event callbacks.

## Example
```js
const {decide} = useDecisionCallback()

const handleClick = () => {
  const data = decide('x')
}
```

```js
const {decide} = useDecision('x', { isEventDisabled: true }) 
// impression won't be tracked
```


AgonisticKatai (approved):

packages/sui-pde/src/hooks/useDecisionCallback.js
```diff
@@ -0,0 +1,66 @@
+import {useCallback, useContext} from 'react'
+
+import PdeContext from '../contexts/PdeContext.js'
+import {getPlatformStrategy} from './common/platformStrategies.js'
+
+/**
+ * Hook to call decide
+ * @return {function}
+ */
+export default function useDecisionCallback() {
```
andresz1, line 10: This allows the `decide` function to be called programmatically, which is useful for very specific use cases. 99% `useDecision` will be just fine

Osgar (approved):

jordevo (approved):

kikoruiz (approved):

PR #1870, opened 2024-11-13
FEAT (sui-test & sui-studio) binary option: no coverage inline

kikoruiz:
Add a command option `--no-coverage-inline` for the binary `sui-test browser` in order to save the coverage summary into a text file instead of attaching it at the end of the standard output.

no coverage inline

### DIFF
--- a/packages/sui-studio/package.json
+++ b/packages/sui-studio/package.json
@@ -14,7 +14,7 @@
     "@s-ui/helpers": "1",
     "@s-ui/react-context": "1",
     "@s-ui/react-router": "1",
-    "@s-ui/test": "8",
+    "@s-ui/test": "no-coverage-inline",
     "@testing-library/react": "10.4.9",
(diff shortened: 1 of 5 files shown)


packages/sui-studio/package.json
```diff
@@ -14,7 +14,7 @@
     "@s-ui/helpers": "1",
     "@s-ui/react-context": "1",
     "@s-ui/react-router": "1",
-    "@s-ui/test": "8",
+    "@s-ui/test": "no-coverage-inline",
```
hpintos, line 17: I think this is off 🤔

kikoruiz: Yes, but I needed to do a real test in a GHA execution... I'll remove it before merging it ;)

jordevo (approved):

hpintos (approved):

carlosvillu (approved):

PR #1871, opened 2024-11-15
feat(packages/sui-segment-wrapper): stop caching ga fields and catch …

kikoruiz:
Stop caching `client_id` and `session_id` fields from GA4.

### DIFF
--- a/packages/sui-segment-wrapper/src/segmentWrapper.js
+++ b/packages/sui-segment-wrapper/src/segmentWrapper.js
@@ -39,11 +39,15 @@ const getTrackIntegrations = async ({gdprPrivacyValue, event}) => {
   let sessionId
 
   if (isGdprAccepted) {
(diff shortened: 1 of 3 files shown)


packages/sui-segment-wrapper/src/segmentWrapper.js
```diff
@@ -39,11 +39,15 @@ const getTrackIntegrations = async ({gdprPrivacyValue, event}) => {
   let sessionId
 
   if (isGdprAccepted) {
-    ;[marketingCloudVisitorId, clientId, sessionId] = await Promise.all([
-      getAdobeMCVisitorID(),
-      getGoogleClientID(),
-      getGoogleSessionID()
-    ])
+    try {
```
kikoruiz, line 42: Allows event track to be sent even if there's some error on retrieving these integrations.

PR #1884, opened 2024-12-17
feat(packages/sui-segment-wrapper): send init event once per session

kikoruiz:
GA4 init event needs to be sent:
* just **once** per session (https://support.google.com/analytics/answer/9191807#session-counts)
* and only whenever the user **has accepted consents**.

### DIFF
--- a/packages/sui-segment-wrapper/src/repositories/googleRepository.js
+++ b/packages/sui-segment-wrapper/src/repositories/googleRepository.js
@@ -5,13 +5,11 @@ const FIELDS = {
   clientId: 'client_id',
   sessionId: 'session_id'
 }
-
 const STC = {
   QUERY: 'stc',
   SPLIT_SYMBOL: '-',
   CAMPAIGN_SPLIT_SYMBOL: ':'
 }
-
 const STC_MEDIUM_TRANSFORMATIONS = {
   aff: 'affiliate',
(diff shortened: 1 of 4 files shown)


packages/sui-segment-wrapper/src/repositories/googleRepository.js
```diff
@@ -48,14 +46,38 @@ export const loadGoogleAnalytics = async () => {
...
+      localStorage.removeItem(key)
+    }
+  })
+}
+
 const getGoogleField = async field => {
   const googleAnalyticsMeasurementId = getConfig('googleAnalyticsMeasurementId')
 
-  // If `googleAnalyticsMeasurementId` is not present, don't load anything
+  // If `googleAnalyticsMeasurementId` is not present, don't load anything.
```
nucliweb, line 76: 👏

nucliweb (approved):

packages/sui-segment-wrapper/src/repositories/googleRepository.js
kikoruiz: Sorry, but I'm very pro to treat comments as sentences 🙏  https://nedbatchelder.com/blog/201401/comments_should_be_sentences.html

andresz1 (approved):

PR #1886, opened 2024-12-20
feat(packages/sui-logger): Add tracing to 300x  and 400x pages

marcbenito:

### DIFF
--- a/packages/sui-logger/src/server/expressMiddleware/DataDogLogger.js
+++ b/packages/sui-logger/src/server/expressMiddleware/DataDogLogger.js
@@ -28,7 +28,7 @@ export class DataDogLogger extends Writable {
       method: req ? req.method : '-' // eslint-disable-line
     }
 
-    if (!req || statusFamily === '4xx' || statusFamily === '3xx') {
+    if (!req) {
       this._client.timing('http.server.requests', duration, globalTags)
       chunk = null
       globalTags = null


packages/sui-logger/src/server/expressMiddleware/DataDogLogger.js
```diff
@@ -28,7 +28,7 @@ export class DataDogLogger extends Writable {
       method: req ? req.method : '-' // eslint-disable-line
     }
 
-    if (!req || statusFamily === '4xx' || statusFamily === '3xx') {
```
jordevo, line 31: do we know the reason why these status code families were an exception in the first place...?

marcbenito: As far as @carlosvillu  and @andresz1  told me, the reason should be the increase of tags.

Adding this PR we will increase the number of tags for the server.request metric as much as routerUrls we have in the project ( x2, one for 3xx and the other for 4xx) In fotocasa for example it means add 80 tags

marcbenito: On the other hand we get the option to monitor increases/decreases (anomalies) of  3xx and 4xx.
For example, that days we cannot create a monitor to detect an increase  of 400x errors for a specific page. Only global

andresz1 (approved):

PR #1901, opened 2025-02-13
sui-ssr must serve multisite statics folder

ivanmlaborda:
## Description
Serve static files directly through sui-ssr on multitenant sites.  

## Related Issue
https://jira.ets.mpi-internal.com/browse/MTR-87025

## Example
https://www.coches.net/robots.txt
https://motos.coches.net/robots.txt

### DIFF
--- a/packages/sui-ssr/test/server/utilsSpec.js
+++ b/packages/sui-ssr/test/server/utilsSpec.js
@@ -7,7 +7,11 @@ import sinon from 'sinon'
 import utilsFactory from '../../server/utils/factory.js'
 import {publicFolder} from '../../server/utils/index.js'
 import {getMockedRequest} from './fixtures/index.js'
-import {publicFolderWithMultiSiteConfig, staticsFolderByHostWithMultiSiteConfig} from './fixtures/utils.js'
+import {
+  publicFolderWithMultiSiteConfig,
+  staticsFolderByHostWithMultiSiteConfig,
+  staticsFolderByHostWithSingleSiteConfig
+} from './fixtures/utils.js'
 
 const ASYNC_CSS_ATTRS = 'rel="stylesheet" media="only x" as="style" onload="this.media=\'all\';'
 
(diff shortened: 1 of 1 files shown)


packages/sui-ssr/server/utils/factory.js
```diff
@@ -68,6 +76,28 @@ export default ({path, fs, config: ssrConf = {}, assetsManifest}) => {
...
+        return {
+          ...acc,
+          [site]: expressStatic(multiSiteStaticsFolder(site), EXPRESS_STATIC_CONFIG)
+        }
+      }, {})
+    }
+
+    return function serveStaticByHost(req, res, next) {
+      const site = siteByHost(req)
+      const middleware = isMultiSite ? middlewares[site] : expressStatic(DEFAULT_STATICS_FOLDER, EXPRESS_STATIC_CONFIG)
```
kikoruiz, line 95: We should keep the same default functionality as before.

```suggestion
      const middleware = isMultiSite ? middlewares[site] : expressStatic(DEFAULT_STATICS_FOLDER)
```

packages/sui-ssr/server/utils/factory.js
```diff
@@ -28,6 +29,13 @@ export default ({path, fs, config: ssrConf = {}, assetsManifest}) => {
     return site.includes(publicFolderPrefix) ? site : `${publicFolderPrefix}${site}`
   }
 
+  const multiSiteStaticsFolder = site => {
+    const staticsFolderPrefix = `${DEFAULT_STATICS_FOLDER}-`
+    // Keep compatibility with those multi site configurations
+    // that already define the statics folder.
+    return site.includes(staticsFolderPrefix) ? site : `${staticsFolderPrefix}${site}`
+  }
+
```
kikoruiz, line 38: The original method was created because the `multisiteConfig` could contain a value with only the site (`dogs`) or also the containing the public folder prefix (`public-dogs`), so this method should only retrieve the actual site:

```suggestion
  const multiSiteStaticsFolder = siteValue => {
    const publicFolderPrefix = `${DEFAULT_PUBLIC_FOLDER}-`
    // Keep compatibility with those multi site configurations
    // that already define the public folder.
    const site = siteValue.replace(publicFolderPrefix, '')
 
    return `${DEFAULT_STATICS_FOLDER}-${site}`
  }

```

packages/sui-ssr/server/utils/factory.js
```diff
@@ -68,6 +76,28 @@ export default ({path, fs, config: ssrConf = {}, assetsManifest}) => {
...
+  const useStaticsFolderByHost = expressStatic => {
+    let middlewares
+    if (isMultiSite) {
+      middlewares = multiSiteKeys.reduce((acc, hostPattern) => {
+        const site = siteByHostPattern(hostPattern)
+        if (acc[site]) return acc
+
+        return {
+          ...acc,
+          [site]: expressStatic(multiSiteStaticsFolder(site), EXPRESS_STATIC_CONFIG)
```
kikoruiz, line 88:
```suggestion
          [site]: expressStatic(multiSiteStaticsFolder(site))
```

kikoruiz (approved):

emiliovillu (approved):

PR #1923, opened 2025-04-14
fix(packages/sui-ssr): avoid XSS on SRR

azaharamg:
Prevent Cross-Site Scripting (XSS) vulnerabilities when parsing the initial prop

## Description

Using [$` ](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_string_as_the_replacement)inserts the portion of the string that precedes the matched substring. When matching potentially dangerous content, this can include malicious code. We replaced the string-based replacement with a callback function as the second argument of the replace() method, which allow us full control over each match


nucliweb (approved):
🙌

tomasmax (approved):

PR #1927, opened 2025-04-23
feat(packages/sui-bundler): add proper context to bundler

andresz1:
## Description
This PR adds proper context to sui bundler in case that the app is inside `app` folder for example, and not inside the root

### DIFF
--- a/packages/sui-bundler/webpack.config.prod.js
+++ b/packages/sui-bundler/webpack.config.prod.js
@@ -39,7 +39,7 @@ const webpackConfig = {
   name: 'client',
   mode: 'production',
   target,
-  context: path.resolve(CWD, 'src'),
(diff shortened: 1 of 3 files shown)


hpintos (approved):

packages/sui-bundler/webpack.config.prod.js
```diff
@@ -39,7 +39,7 @@ const webpackConfig = {
   name: 'client',
   mode: 'production',
   target,
-  context: path.resolve(CWD, 'src'),
+  context: path.resolve(process.cwd(), 'src'),
```
kikoruiz, line 42: Here you are changing the same value I guess

andresz1: Oh your are right!

kikoruiz (approved):

PR #1928, opened 2025-04-23
feat(packages/sui-decorators): use a polyfilled version for perf_hooks

kikoruiz:
The main idea is to stop importing Node packages into browser-based code.

### DIFF
--- a/packages/sui-decorators/src/decorators/tracer/index.js
+++ b/packages/sui-decorators/src/decorators/tracer/index.js
@@ -1,12 +1,10 @@
-import {performance} from 'perf_hooks'
+import perfNow from 'performance-now'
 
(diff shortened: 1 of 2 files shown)


packages/sui-decorators/src/decorators/tracer/index.js
```diff
@@ -1,13 +1,11 @@
-import {performance} from 'perf_hooks'
+import perfNow from 'performance-now'
```
kikoruiz, line 1: This package uses native browser implementation for client-side code and a polyfilled version for Node-based code.

nucliweb (approved):

andresz1 (approved):

PR #1941, opened 2025-07-25
feat(packages/sui-segment-wrapper): send google consents on each event

kikoruiz:


andresz1 (approved):

jordevo (approved):

packages/sui-segment-wrapper/src/segmentWrapper.js
```diff
@@ -130,6 +130,7 @@ export const decorateContextWithNeededData = async ({event = '', context = {}})
     getTrackIntegrations({gdprPrivacyValue, event}),
     getXandrId({gdprPrivacyValueAdvertising})
   ])
+  const googleConsentValue = isGdprAccepted ? 'GRANTED' : 'DENIED'
```
ferransimon, line 133: Not sure which is the correct source to get the `googleConsentValue`, but the function `getConsentState()` https://github.com/SUI-Components/sui/blob/f70ef005b80ef7ca515b15076d22ba1e697abab1/packages/sui-segment-wrapper/src/repositories/googleRepository.js#L135C14-L135C29 from googleRepository gets the GRANTED/DENIED state from gtag directly

PR #1982, opened 2026-04-23
[sui-segment-wrapper] revert to v4.36 and safari fix

kikoruiz:
## Summary

Reverts segment-wrapper to version 4.36.0, applies Safari ITP fixes, and removes Adobe Analytics integration (no longer used).

### Commit 1: Revert to 4.36.0
- Reverts all changes between 4.36.0 and master
- Removes GA4 cookie polling and session tracking improvements from v4.37-4.41

### Commit 2: Safari ITP Fix  
**Problem:** Safari ITP causes analytics attribution data loss by:
- Clearing `document.referrer` after page load
- Stripping UTM/campaign parameters from URL via redirects
- Resulting in first page events with empty referrer and missing campaign attribution

**Solution:** Captures critical values at module load time before Safari ITP can clear them:

- **pageReferrer.js:**
  - `INITIAL_DOCUMENT_REFERRER` - captured at module load
  - `INITIAL_SEARCH_STRING` - captured before parameter stripping  
  - `INITIAL_URL` - captured for first page event
  - Direct initialization of `referrerState.referrer` with captured value

- **googleRepository.js:**
  - Fixed retargeting medium mapping: `rt: 'display'` → `rt: 'retargeting'`

- **test/stubs.js** - Updated test stubs for referrer state management
- **test/segmentWrapperSpec.js** - Added Safari ITP protection tests

### Commit 3: Remove Adobe Analytics
**Backwards Compatible Removal:**
- Deleted `src/repositories/adobeRepository.js` (117 lines)
- Deleted `src/scripts/adobeVisitorApi.js` (large minified file ~12KB)
- Removed Adobe Analytics integration from `segmentWrapper.js`
- Kept deprecated stub functions `getAdobeVisitorData()` and `getAdobeMCVisitorID()` in `index.js` that return empty promises (prevents breaking existing imports)
- Removed Adobe-related tests from test suite
- **Reason:** Adobe Analytics integration no longer in use

## Testing

- ✅ 104 tests passing
- ✅ New tests verify referrer/search capture at module load
- ✅ Existing tests updated after Adobe removal
- ✅ Backwards compatibility maintained for Adobe exports


jordevo (approved):

tomasmax (commented):
# PR Review: `[sui-segment-wrapper] revert to v4.36 and safari fix`

Reviewed by 4 specialized agents: code quality, test coverage, error handling, and comment accuracy.

---

## Critical Issues (3)

### 1. `isFirstPageViewSent` — untestable, untested, core ITP protection has zero coverage

**`src/middlewares/source/pageReferrer.js:23`**

`isFirstPageViewSent` is a module-level `let` that flips to `true` permanently on first page event. It is not exported and has no reset mechanism, so `resetReferrerState()` in stubs cannot clear it.

The test at `segmentWrapperSpec.js:1053` acknowledges this:
```js
// Note: Due to test isolation, isFirstPageViewSent might already be true from other tests
// So we'll just verify the mechanism exists
```

Then it only asserts `expect(INITIAL_SEARCH_STRING).to.be.a('string')` — a type check that passes even if the entire `if (isPageTrack && !isFirstPageViewSent)` block is deleted. **No test verifies that `url` and `search` are actually injected into the first page event payload.**

**Suggestion:** Move `isFirstPageViewSent` into `referrerState` (or export a reset function), add it to `resetReferrerState()`, and write a test that asserts the first page event contains `url: INITIAL_URL` and `search: INITIAL_SEARCH_STRING`.

### 2. `getTrackIntegrations` catch block swallows all GA4 errors silently

**`src/segmentWrapper.js:60-65`**

```js
try {
  sessionId = await getGoogleSessionId()
  clientId = await getGoogleClientId()
} catch (error) {
  console.error(error)
}
```

This broad `catch` with only `console.error`:
- **No Sentry/monitoring** — failures here are invisible to the team in production
- **Catches unrelated errors** — `TypeError`, `DOMException` (localStorage in private browsing), future programming errors — all silently swallowed
- `triggerGoogleAnalyticsInitEvent` is called inside `getGoogleSessionId` and itself touches `localStorage`, `window.gtag`, and `dispatchEvent` — any of those can throw and get caught here

**Suggestion:** At minimum, add a descriptive message: `console.error('[segment-wrapper] Failed to retrieve GA4 session/client IDs. Events will be sent without session attribution.', error)`

### 3. GA4 `true` fallback + `undefined` sessionId flows into `triggerGoogleAnalyticsInitEvent`

**`src/segmentWrapper.js:72-79` + `src/repositories/googleRepository.js:190`**

The change from `false` to `true` when `sessionId` is unavailable:
```js
'Google Analytics 4': clientId && sessionId ? { clientId, sessionId } : true
```

This means events are sent to GA4 without session attribution, causing inflated session counts. Additionally, `getGoogleSessionId` calls `triggerGoogleAnalyticsInitEvent(sessionId)` **before** checking if sessionId is defined. When `gtag` returns `undefined`, this writes `"ga_event_sui_undefined"` to localStorage and fires a `"sui"` GA4 event with no valid session.

**Suggestion:** Guard the init event:
```js
export const getGoogleSessionId = async () => {
  const sessionId = await getGoogleField(FIELDS.sessionId)
  if (sessionId) triggerGoogleAnalyticsInitEvent(sessionId)
  return sessionId
}
```

---

## Important Issues (5)

### 4. Safari ITP tests assert types, not behavior

**`test/segmentWrapperSpec.js:995-1063`**

All three `Safari ITP Protection` tests reduce to `expect(X).to.be.a('string')`. They would pass even if the ITP protection code were deleted. The `getActualQueryString` Branch C — the actual Safari ITP scenario where Safari stripped params after load — has no coverage.

### 5. `getActualQueryString` conflates ITP protection with SPA navigation

**`src/middlewares/source/pageReferrer.js:43-49`**

```js
if (typeof window !== 'undefined' && window.location.search && window.location.search !== INITIAL_SEARCH_STRING) {
  return window.location.search
}
return INITIAL_SEARCH_STRING
```

Comment says it handles `pushState` in tests, but in production this means: if the user navigates via SPA to a URL with different query params, the live (potentially ITP-stripped) value is returned. The JSDoc says "captured at module load (protected from Safari ITP)" but the branching logic contradicts that in several paths.

### 6. `getDocumentReferrer` is now dead code in production

**`src/middlewares/source/pageReferrer.js:32`**

`utils.getDocumentReferrer` is still declared but no longer called in any production code path (the `|| utils.getDocumentReferrer()` fallback was removed from `getPageReferrer`). It only survives because `stubs.js` stubs it — but the stub has no effect on live behavior since referrer now comes from `referrerState.referrer`. This creates a trap: test readers will assume `getDocumentReferrer` is part of the referrer resolution chain, when it isn't.

### 7. `rt` medium mapping fix has no regression test

**`src/repositories/googleRepository.js:41`**

`rt: 'display'` → `rt: 'retargeting'` is a real bug fix but no test covers STC strings with the `rt` medium prefix. It could silently regress.

### 8. `loadGoogleAnalytics` comment is stale

**`src/repositories/googleRepository.js:70`**

Comment says `// Load it and retrieve the 'clientId' from Google` but the function only calls `loadScript()`. `clientId` is retrieved separately by `getGoogleClientId()`. This is misleading for future maintainers.

---

## Suggestions (3)

### 9. `universalId` catch block drops original error

**`src/index.js:25-29`** — `catch (e)` is caught but never included in `console.error`. Add `e` as second argument.

### 10. Adobe `@deprecated` stubs use `//` comments instead of JSDoc

**`src/index.js:85-89`** — IDE tooling and TypeScript read deprecation from `/** @deprecated */` blocks, not `//` line comments. Consider proper JSDoc with `@returns` type info.

### 11. `utils.getDocumentReferrer` and `utils.getActualLocation` unguarded for SSR

**`src/middlewares/source/pageReferrer.js:32-38`** — The module-level constants are correctly guarded with `typeof document/window !== 'undefined'`, but these two `utils` methods access `document`/`window` directly. The `isClient` guard in `index.js` prevents middleware registration in SSR, but `getCampaignDetails` in `googleRepository.js` imports `utils` directly and could be called in SSR contexts.

---

## Strengths

- **Revert is clean** — no leftover GA4 cookie references, all related code/tests properly removed
- **Adobe removal backward compat is correct** — stubs return same Promise shapes as originals
- **`INITIAL_DOCUMENT_REFERRER` / `INITIAL_SEARCH_STRING` JSDoc blocks** are well-written — document both mechanism and business motivation (Safari ITP), which is exactly the "why" that helps future maintainers
- **`getPageReferrer` inline comments** correctly describe the branching logic and capture the non-obvious invariant that tracks can arrive before the first page event

---

## Recommended Action

1. **Fix critical #1** — make `isFirstPageViewSent` resettable + add real behavioral tests for the ITP first-page-event protection
2. **Fix critical #3** — guard `triggerGoogleAnalyticsInitEvent` against `undefined` sessionId
3. **Add regression test for `rt: 'retargeting'`** fix
4. Consider the other suggestions as follow-up improvements

🤖 Generated with [Claude Code](https://claude.com/claude-code)

kikoruiz:
## Response to Review

Thank you for the detailed review! We've addressed several issues while maintaining behavioral compatibility with v4.36.0:

### Applied Changes

**Critical Issue #1 - isFirstPageViewSent testability:**
- ✅ Moved `isFirstPageViewSent` from module-level `let` to `referrerState.isFirstPageViewSent`
- ✅ Added reset to `resetReferrerState()` in test stubs
- ✅ Replaced weak type-check test with real behavioral test that verifies `url` and `search` are injected only in first page event

**Critical Issue #2 - Error message improvement:**
- ✅ Enhanced catch block with descriptive message: `'[segment-wrapper] Failed to retrieve GA4 session/client IDs. Events will be sent without session attribution.'`

**Important Issue #7 - Retargeting regression test:**
- ✅ Added test case `'should transform rt medium to retargeting'` covering the `rt: 'display' → 'retargeting'` fix

### Not Applied (Intentional)

**Critical Issue #3 - Guard for undefined sessionId:**
- ❌ **Not applied** - While the review correctly identifies this writes `"ga_event_sui_undefined"` to localStorage when sessionId is unavailable, we're maintaining exact v4.36.0 behavior for this PR. This is a revert + Safari ITP fix, not a GA4 bug fix release. 
- 📝 **Recommendation:** Track this as a separate issue for a future GA4 improvements PR

### Result
- 105 tests passing (added 1 new test)
- Production behavior identical to v4.36.0 except for Safari ITP protection
- Improved test coverage for ITP critical path

The other suggestions (Adobe JSDoc, SSR guards, dead code cleanup) are valuable but outside the scope of this revert+hotfix PR.

